PR #1, opened 2016-01-08
Migrate Google Directory Service Test and Turn on Travis

xwsxethan:
[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/1)


blueandgold:
Cool, this is a good start on something that is independent of flask.  Keep in mind there will be other tests  where we will need to learn how to handle flask.  :+1:

jpevarnek:
I'm not going to get a chance to review this until later tonight, sorry
about that

@henryc http://flask.pocoo.org/docs/0.10/testing/ has more information on
testing Flask

xwsxethan:
@jpevarnek That's fine. It's no rush, I just wanted to get the ball rolling with something (reasonably) simple as well as turning on travis so I can what works and what doesn't. I'll start looking at testing flask in another PR. Thanks!

jpevarnek:
Looking at these tests in general, they seem to be incredibly coupled to the implementation.

A couple examples:
`testInit` checks that the constructor initializes a property (that really should not be public if Python has a concept of private elements) using a very precise call.  In my view of what the tests should cover, I should be able to substitute an implementation of the class that 

`testConstants` verifies the value of some hardcoded constants in the file.  I am actually unable to imagine a case where this test would fail other than from me changing the constants intentionally...

`testGetUsers` verifies that the `GetUsers` method is implemented precisely by a function call that looks like
```
      request = self.service.users().list(customer=MY_CUSTOMER_ALIAS,
                                          maxResults=500,
                                          pageToken='',
                                          projection='full',
                                          orderBy='email')
```
, again, looking at the precise implementation.  I should be able to substitute this for a version that does
`self.service.users().list(customer=ACTUAL_CUSTOMER_ID, maxResults=20, pageToken=blah, projection='custom', customFieldMask='only,what,I,need',orderBy='givenName')`, call that a couple times until I get all the users, and then return the result.  I think it would be far more useful to have multiple tests at this function that verify the output and not exactly the way it is implemented.  It may also be useful to generate a full mock of the service that we can use in an easier way that setting individual parameters each time.

I've held off on looking at every single method for now as I think it would be more useful to discuss the testing methodology as a broad concept first.

Sorry for the delay in getting to this review

---

Reviewed 2 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[.travis.yml, line 7 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/1#-K7qNvXRY0gD0vBE9BAt:-K7qNvXRY0gD0vBE9BAu:-503550308) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a29cb1b554c6e33e3a2a10761f22b5c938ffa0b2/.travis.yml#L7)):*
I think you can replace the first four lines with `pip install -r requirements.txt`

---

*[ufo/google_directory_service_test.py, line 4 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/1#-K7qUHw9Zf-NDmaDucwh:-K7qUHw9Zf-NDmaDucwi:1284169293) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a29cb1b554c6e33e3a2a10761f22b5c938ffa0b2/ufo/google_directory_service_test.py#L4)):*
My reading of the style guide seems to indicate that we should be using imports for modules and packages only (not constants)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/1#-:-K7qVVplrpAalG-lAXBh:1007262491)*

xwsxethan:
I think you may have deleted something in your explanation for testInit at the end, but overall I understand what you're saying, and I agree, at the moment it is very coupled. It is much more useful to test and assert on real data rather than just asserting that foo mock calls bar mock (which is really just saying that things are wired up as written in the code). When doing these tests a while back, I was going for some baseline of testing that we could iterate on without having to manually verify calls in the code, but I got a bit carried away as you can see, and we haven't updated them since due to other priorities.

In spite of that, I will say that having an assert based on your constants can be valid for exactly the case you pointed out. If all the application and test code references a constant, the constant could be changed with all tests still passing (as the test references the same new value as the application). The reason for having a test for this (even though it seems very trivial) is to catch when the constant is changed either intentionally or otherwise, such as a typo, since all other tests could pass while it potentially wreaks havoc on the deployed application. In this way it forces you to update the test when you make that change to verify that you really did want to change its value rather than the change being a typo.

To improve these overall, I suggest we create a test domain that we maintain and has certain properties (users, groups, etc.) that we can assert are found in the calls. This should probably be separate from our existing henrychang.mygbiz domain since that is modified occasionally. It would allow us to actually call through to the api with real queries and see that we receive valid results.

The alternative to this, as you pointed out, would be to mock out the entire service, which is more or less what I've done in some of these calls. I can expand that and replace it in the init function (whose test can likely be dropped since it is tested implicitly anyway). From there, each test would just assert that it receives and parses out values from the users and groups we defined in the mock.

Either approach will work, it just comes down to how focused we want the individual tests to be. For the moment, I suggest perhaps I remove the google_directory_service_test.py file from this PR and submit with the changes to get Travis CI up and running. Whatever approach we decide on, I can rewrite google_directory_service_test.py and resubmit it in another PR as appropriate so we aren't blocked on using Travis any longer.

---

Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[.travis.yml, line 7 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/1#-K7qNvXRY0gD0vBE9BAt:-K7qpilMJJEUniBwDOSH:-1358906501) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a29cb1b554c6e33e3a2a10761f22b5c938ffa0b2/.travis.yml#L7)):*
Which first four lines do you mean here, like replace sudo: false, language, etc. or pip install virtualenv, virtualenv venv, etc. ? I set it up based on what was in our email chain for how to run locally.

---

*[ufo/google_directory_service_test.py, line 4 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/1#-K7qUHw9Zf-NDmaDucwh:-K7qpAJeBwhYZCls4OMJ:464487789) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a29cb1b554c6e33e3a2a10761f22b5c938ffa0b2/ufo/google_directory_service_test.py#L4)):*
In general I would agree with that, but this was more of an optimization for space so everything doesn't go over 80 characters since google_directory_service is so long. Henry's approach thus far has been for more descriptive names, which means I haven't imported this as gds or something similar so this was the compromise. Does this seem reasonable then?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/1#-:-K7qqPOinJDSivwiPuqi:-307498039)*

goldspin:
Thanks Jonathan for the thorough review!  So I just want to add my 2 cents here as well, and we can try to converge later.

So, typically, I always though there are primarily 2 reasons for testing:
#1 verify our logic statements are correct
#2 safeguard our app from breaking, i.e. when someone makes a typo, or missed a parameter to a method call
That's why I think it's still worthwhile to assert on our actual implementation and api calls.

Ideally as you mentioned, we should test in a more functional fashion.  But there is no way to start a real directory server inside our unit tests.  So the best that we can do is to mock out the api calls and verify they are called correctly. 

Where we can do functional testing is with webdriver.  We already have some in place, and just need to extend those for more coverage.  And Ethan's ideal to have a dedicated domain for testing is great.

---

Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/1#-:-K7r7n09Jbvwc3ZO9xPY:-2092653126)*

jpevarnek:
I want to step back a bit and clarify what I think the purpose of a test should (and should not) be, just so we're sure we're on the same page here.

What a test should catch: the thing I care about for an interface breaking

What a test should not catch: The complete re-implementation of a method.  If an implementation detail is changed (concrete examples in this case: switching the number of things batched together in an intermediary state, when an internal variable is initialized (when needed vs. in the constructor)), the test should not have to be changed at all.  In the case where an underlying library is changed (e.g. moving from a google-written api to one provided by a third party), the only thing that should need to change

My way of thinking about tests: "Given that I have no code written, what do I want to verify will happen once I write the code?" (lookup: test driven development).  If I need to specifically worry about implementation details of the class when writing the test, I probably made a mistake somewhere.  When writing the test, other than maybe setting up infrastructure, I should not have to read anything outside of a .h file (or, in python, generated pydocs)

With that being said, it may be useful to get a consensus on what the tests we have here should cover going forward.

A couple things I'm thinking here that may help with any discussion:

My opinion is that an accidental change made to an implementation that affects behaviour in a non-breaking (breaking in this case defined as impacting the things we care about) way should be caught by a code review, not by a test.  If the test is catching that, my instinct is that it's checking too much.

One of the suggestions above is to create a test domain and use it.  I think that's a great idea and I think we should definitely include some integration tests like that in the framework.  I do not, however, think that that affects the usefulness of tests on the values of constants or implementation details within a function.

I think a lot of the problem here is my failure to understand the problems you are trying to catch with these tests, it might be useful if you could clarify what some of the errors they would catch (that would not be caught by output-based tests) would be.

Thanks a lot for taking the time to discuss this!

---

Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/1#-:-K7rKoMnBMOpMD5cizfS:2050998880)*

jpevarnek:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[.travis.yml, line 7 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/1#-K7qNvXRY0gD0vBE9BAt:-K7rU9maUbM1xJTGackU:1115952484) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a29cb1b554c6e33e3a2a10761f22b5c938ffa0b2/.travis.yml#L7)):*
```
pip install virtualenv
virtualenv venv
.  venv/bin/active
pip install -r requirements.txt
```

can become

`pip install -r requirements.txt`

Travis will handle the creation of a virtual environment for you.

---

*[ufo/google_directory_service_test.py, line 4 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/1#-K7qUHw9Zf-NDmaDucwh:-K7rURMfiDEAIZF41YlD:-1565296012) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a29cb1b554c6e33e3a2a10761f22b5c938ffa0b2/ufo/google_directory_service_test.py#L4)):*
I'd still be in favor of following the style guide precisely for this (i.e. `import google_directory_service` and accessing as `google_directory_service.MY_CUSTOMER_ALIAS` where needed (though, as noted above, I'd disagree about that constant ever being needed in the test).

For line length, I'd say just break it up as needed.

I guess we never actually decided upon a style guide to follow for this project though...  If you want, I'd be fine with forking https://google.github.io/styleguide/pyguide.html and making changes to that as necessary for this project if you think there are problems with it since we do not have actual constraints on us.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/1)*

xwsxethan:
I think we should definitely continue this conversation so I don't want to derail it, but perhaps lets move it to the email thread we already have. In the interest of proceeding forward with getting our testing setup, I've gone ahead and removed google_directory_service_test from this PR and will resubmit with whatever changes we decide to follow. Please thumbs up this PR if the changes to turn on Travis look correct.

---

Review status: 2 of 3 files reviewed at latest revision, all discussions resolved.

---

*[.travis.yml, line 7 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/1#-K7qNvXRY0gD0vBE9BAt:-K7robMZMZirtH_bwANa:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a29cb1b554c6e33e3a2a10761f22b5c938ffa0b2/.travis.yml#L7)):*
Done.

---

*[ufo/google_directory_service_test.py, line 4 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/1#-K7qUHw9Zf-NDmaDucwh:-K7rod8epINLRvymBB6X:210726495) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a29cb1b554c6e33e3a2a10761f22b5c938ffa0b2/ufo/google_directory_service_test.py#L4)):*
I'll look at this in whatever the new PR ends up being.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/1#-:-K7rp8Eerx4AbEeTOEGh:688999896)*

jpevarnek:
:lgtm: (and let's talk about the testing mentality offline)

---

Review status: 1 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/1#-:-K7s0XZWk92jgj28UZEp:1488243084)*

PR #2, opened 2016-01-12
Restructure the database code.

blueandgold:
This PR accomplishes the following:
- move the direct database operations out of the handlers
- create a base class object that is used declaratively, and common operations
- fixed the commitDB() being called mulitple times, on all requests

The most important file is database.py.  Before looking at that code, please look through the sqlalchemy documentation at below.
http://flask.pocoo.org/docs/0.10/patterns/sqlalchemy/#declarative

I hope that I didn't leave any loose ends, but if anything gets through or breaks, please let me know.

### DIFF
--- a/ufo/__init__.py
+++ b/ufo/__init__.py
@@ -26,7 +26,6 @@ db = sqlalchemy.SQLAlchemy(app)
 
 # DB needs to be defined before this point
 import database
-import models
 from models import Config
 
 def get_user_config():
--- a/ufo/__init__.py
+++ b/ufo/__init__.py
@@ -26,7 +26,6 @@ db = sqlalchemy.SQLAlchemy(app)
 
 # DB needs to be defined before this point
 import database
-from database import db_session
 import models
 from models import Config
 
@@ -57,4 +56,4 @@ import routes
 
 @app.teardown_appcontext
 def shutdown_session(exception=None):
-    db_session.remove()
+    database.db_session.remove()
--- a/ufo/__init__.py
+++ b/ufo/__init__.py
@@ -1,8 +1,7 @@
-import os
 import flask
-import functools
-
 from flask.ext import sqlalchemy
+import functools
+import os
 
 app = flask.Flask(__name__, instance_relative_config=True)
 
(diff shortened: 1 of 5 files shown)


ufo/__init__.py
```diff
@@ -26,30 +25,17 @@
...
-  if credentials is not None:
-    config = get_user_config()
-    json_credentials = credentials.to_json()
-    if config.credentials != json_credentials:
-      config.credentials = json_credentials
-
-  db.session.commit()
-
-  return response
+from models import Config
```
xwsxethan: Do you need this specifically since models is already imported overall? If we're not referencing models anymore, then maybe remove that and only have this import.

ufo/proxy_server.py
```diff
@@ -5,6 +5,8 @@
 import logging
 import flask
 import models
+import database
+from models import ProxyServer
```
xwsxethan: Same comments here about the imports overlapping.

xwsxethan:
Looks great overall. Just a couple imports to clean up, but otherwise :+1:

blueandgold:
Done.  Good catch on the overlapping imports.

@jpevarnek, please take a look here if you have a chance.  But I have to submit first, so that ethan can merge in his changes to make the app testable.

jpevarnek:
Hey there, couple things I wanted to check in about real quick on this:

So, I realize that I had the import order wrong originally already, but in `__init__.py`, I'm pretty sure that the order should have been functools, os, flask, flask.ext.sqlalchemy

`from models import Config` again seems to violate the Python style guide I'm looking at (https://google.github.io/styleguide/pyguide.html#Imports), I think we still need to clarify what guide we are using though.

`database.GetFirstRecord` does not return the first record.  Implementation for that would be `model.query.order_by({PRIMARY KEY OF MODEL}).limit(1).all()`.  We should either fix the function name there or fix the implementation.  Personally, I'd just recommend deleting that function.

It looks like indentation here was varied, I'm assuming that's because I screwed up and used two-space indentation, can you confirm that?

Can you explain the reasoning for creating the separate database model and the reasoning for writing proxy functions instead of using the class methods directly?  At the moment, it's completely ignoring the configuration values for the database and is switching away from using the Flask-specific version of SQLAlchemy (except in a few places where that is still referenced).

It does not look like there is any code to handle committing the changes to existing models (e.g. changing the config will not take effect unless a new record is added/removed from something later).  Am I missing something?

Thanks!

PR #3, opened 2016-01-13
Enable Landscape.io Health Checks and Set Configs

xwsxethan:
Adding the landscape config and pylintrc file so landscape can better understand our repo. We should be getting more informative results with these.


xwsxethan:
@blueandgold to review as well.

jpevarnek:
:thumbsup: from me

blueandgold:
:+1:

PR #4, opened 2016-01-13
Flask User Test Baseline to Iterate Upon

xwsxethan:
This is the first, most basic test for a handler function that we can iterate upon. It sets up and tears down the database. This only tests the list users function since it was first. In a separate PR, I'll add the other tests once they are migrated, but that should be trivial now that the API is setup.

Explanation of higher level changes:
1. Importing using from . import foo causes a problem with running tests via python -m unittest ... To combat this, I moved everything from **init**.py over to a new file called management_server.py which holds the exact same data.
2. The change to import via a named reference causes a new issue in the setup_database.py file since it doesn't know what to import. Thus, I recreated **init**.py with it simply blank so that python still recognizes the ufo folder as a package with distinct modules as files.
3. The config.py file represents the combination of the previous appengine_config and regular config. I commented out the nonrelevant portions. We can keep this mostly as is in the future or move it to management_server.py if we like. I needed it here in order to keep our PATHS['your-path-here'] references. In a future PR, I'll update the routes on each handler to similarly reference those.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/4)


xwsxethan:
@jpevarnek to review as well.

jpevarnek:
Haven't looked at the actual tests yet, but the rest of my comments are below.  Pretty much, I think I'm mostly confused about the config file.

Please update `Procfile` as well

---

Reviewed 15 of 16 files at r1.
Review status: 15 of 16 files reviewed at latest revision, 6 unresolved discussions.

---

*[.gitignore, line 6 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x5iFMPta8jehxkSX1:-K7x5iFMPta8jehxkSX2:1373419504) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/.gitignore#L6)):*
Can you explain what these entries are?  Thanks.

---

*[ufo/config.py, line 10 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x69L-xlavQfNUvAvM:-K7x69L-xlavQfNUvAvN:-1071524083) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/ufo/config.py#L10)):*
Can you clarify why we are not using `app.jinja_env` here?

---

*[ufo/config.py, line 16 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x6HtDYZDCyiWgyBxs:-K7x6HtDYZDCyiWgyBxt:340998818) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/ufo/config.py#L16)):*
For the paths, I'd personally tend towards using the Flask convention of `flask.url_for` when referring to them instead of having the config array here.

---

*[ufo/config.py, line 46 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x6G9NRPM02F4KgqLP:-K7x6G9NRPM02F4KgqLQ:2135449273) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/ufo/config.py#L46)):*
What is this for?

---

*[ufo/config.py, line 48 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x6ZiLXZsaxVexLMvW:-K7x6ZiLXZsaxVexLMvX:-1375415237) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/ufo/config.py#L48)):*
Should be able to drop this

---

*[ufo/config.py, line 49 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x6dFf8htNQXnDuVto:-K7x6dFf8htNQXnDuVtp:-2034340387) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/ufo/config.py#L49)):*
Can you clarify where all these values are used actually

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-:-K7x5uO6r8dLdHLmul0A:-997435762)*

xwsxethan:
For the config file, I just wanted to bring up the topic of what all we needed there, so thanks for the comments. Most of it can be dropped based on flask equivalents it seems like. I'll update the procfile shortly. Thanks!

---

Review status: 15 of 16 files reviewed at latest revision, 6 unresolved discussions.

---

*[.gitignore, line 6 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x5iFMPta8jehxkSX1:-K7xFZVT_n8Jva3MCtmA:-743097787) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/.gitignore#L6)):*
I'm not entirely sure what .cache is. My assumption is that it is something from the database as it started showing up around that time. If no one else is seeing it when testing locally, then I'll remove it. ufo/venv/* is the same venv for me when I stupidly setup venv in the ufo directory instead of the top level. ufo/test.db is the test database.

---

*[ufo/config.py, line 10 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x69L-xlavQfNUvAvM:-K7xFxCcZVmVU6mvMuys:449803844) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/ufo/config.py#L10)):*
This was a holdover from the old style. We can probably use app.jinja_env instead. Will remove.

---

*[ufo/config.py, line 16 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x6HtDYZDCyiWgyBxs:-K7xGDqXixKPGMsXlhkx:1049951479) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/ufo/config.py#L16)):*
Can that be used in the unit and functional tests as well? The reason we went with this in the other repo was because it could be referenced within application code, unit test code, functional (selenium) test code, and the html via the Jinja globals which we setup at the bottom.

---

*[ufo/config.py, line 46 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x6G9NRPM02F4KgqLP:-K7xGiby1ItUBEOO9w7E:-1730000371) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/ufo/config.py#L46)):*
It was for setting up xsrf and placing the token into the globals so that it can be referenced in the html files. It may no longer be necessary depending on what you modified in the migration.

---

*[ufo/config.py, line 48 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x6ZiLXZsaxVexLMvW:-K7xH6BTXc9aUTMMFgys:-1574472319) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/ufo/config.py#L48)):*
How do we actually get the base url in flask for wherever the server is running?

---

*[ufo/config.py, line 49 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x6dFf8htNQXnDuVto:-K7xHJmnS5kXaWXpfyz8:-1438722792) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/ufo/config.py#L49)):*
The patterns and error messages are used for input validation on polymer input boxes. Specifically, these are used to sanitize the inputs for adding users. Email validation is for validating just an email address specified when manually adding a user (don't think that code made it over in migration as I did it after you prototyped this, see here: https://github.com/uProxy/ufo-management-server/blob/master/templates/add_user.html#L91). Key validation is for an email address or unique key according to google directory service. So for instance when adding users via group, you can specify the group's unique id, which is pure alphanumerics, or you can specify the group email address.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-:-K7xIA6_k43XZ7_YWcHo:2143284371)*

blueandgold:
Nice work overall.  I just have a couple of questions.

#1 Will there be any side-effects by moving out of __init__.py?   From the flask documentation below, it seems like using __init__.py is an important part of the app structure.  (Search for "quick checklist" and the 2 mentioned items.
http://flask.pocoo.org/docs/0.10/patterns/packages/

#2 I am still concerned that we weren't able to find a seam that will allow us to mock out the decorators.  If in the future, we have other decorators (i.e. xsrf or login checks), will we have a way to configure them or mock them?

xwsxethan:
Hmmm, I hadn't seen that before, but I can say that this works. I've booted up the test server locally and everything runs fine.

For our steps, we were already importing code into the top level directory from the lower module (as in, in run.py and setup_database.py we previously did from ufo import app, db or something similar). We're still doing those imports now, but they are named for a module in the directory (from ufo.management_server import app, db). In the run script we still tell it where to find the app to execute. For the code in the lower level, it now imports via a name as well, so instead of from . import app, db, it does from management_server import app, db, which seems to work fine. ¯\_(ツ)_/¯

As for mocking future decorators, it will depend on where we define and import them. My thought is that we should be able to mock those out correctly if they're defined in either management_server.py or some other file, say admin.py as we had before, so long as we don't import the function directly, but that is part of the update we need to do for the style guide anyway. So in user.py for instance, we'll need to change the import to be just import management_server (and not from management_server import setup_required). Doing it that way should allow us to mock them out before importing user in the test just as we had in the other repo. I haven't tried this yet, but hopefully I can flush it out when I address the things Jonathan pointed out in his comments.

---

Review status: 15 of 16 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-:-K8-ddBJxKnOxlolQcSr:-1591515123)*

ufo/management_server.py
```diff
@@ -0,0 +1,58 @@
...
+  """Returns the current user-defined configuration from the database"""
+  config = database.GetFirstRecord(Config)
+  if config is None:
+    config = Config()
+    config.id = 0
+    database.Add(config)
+
+  return config
+
+def setup_required(f):
```
blueandgold: nit: can we change "f" to  at least "func" :)

ufo/user_test.py
```diff
@@ -0,0 +1,85 @@
...
+    # management_server.db.create_all()
+    self.engine = create_engine(self.db_uri, convert_unicode=True)
+    self.session = scoped_session(sessionmaker(autocommit=False,
+                                               autoflush=False,
+                                               bind=self.engine))
+    database.Base.query = self.session.query_property()
+    import models
+    database.Base.metadata.create_all(bind=self.engine)
+
+    self.mock_config = models.Config()
```
blueandgold: nit: because this is not actually a mock, so can we change this to be like below:
self.config = models.Config()

ufo/user_test.py
```diff
@@ -0,0 +1,85 @@
...
+  def testListUsersHandler(self):
+    """Test the list user handler displays users from the database."""
+    fake_users = []
+    for x in range(0, len(FAKE_EMAILS)):
+      fake_user = models.User()
+      fake_user.email = FAKE_EMAILS[x]
+      fake_user.name = FAKE_NAMES[x]
+      fake_users.append(fake_user)
+      self.session.add(fake_user)
+    self.session.commit()
```
blueandgold: This is turning into an integration test with the database, whereas we should try to test the handler code in isolation.  :)  So, instead of adding these to the database to set them up, it might be better to mock out the pending GetALL() that jonathan is working on.

https://github.com/uProxy/ufo-management-server-flask/blob/database-model-methods/ufo/user.py#L103

xwsxethan:
@blueandgold I've made the changes you requested, but I'm waiting on @jpevarnek to submit so I can adjust for the new db calls. Done.

---

Review status: 15 of 16 files reviewed at latest revision, 6 unresolved discussions.

---

*[ufo/management_server.py, line 41 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x4hkawzIjmiUrV1Ne-r1-41:-K80Bk0I1k8qeGqW7TRR:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/ufo/management_server.py#L41)):*
Done.

---

*[ufo/user_test.py, line 42 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x4hkawzIjmiUrV1Nf-r1-42:-K80BlTxZ_0VzRNSd1Vj:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/ufo/user_test.py#L42)):*
Done.

---

*[ufo/user_test.py, line 69 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x4hkawzIjmiUrV1Nf-r1-69:-K80Bnd2fCSmi-pQCCWv:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/ufo/user_test.py#L69)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-:-K80Bq_NH_d4Io7S79Xp:1001092675)*

ufo/user_test.py
blueandgold: Unless Jonathan has any thoughts, would it be possible to do a sanity check with the flask community to make sure this app structure is okay and will not have any subtle framework issues?  As I am also learning flask as well, it would be reassuring that we are not missing something that made "__init__.py" necessary.

blueandgold:
Unless Jonathan has any thoughts, would it be possible to do a sanity check with the flask community to make sure this app structure is okay and will not have any subtle framework issues? As I am also learning flask as well, it would be reassuring that we are not missing something that made "init.py" necessary.

Also regarding decorator mocking, setup_required is in the managment_server.py now.  So, can we test to see if it's possible to mock it out as per your previous comment?
@management_server.setup_required.
I am curious to know if the framework would execute the decorators before the test has a chance to execute.

xwsxethan:
@blueandgold and @jpevarnek please check this again now. I was able to find a test structure that worked using the old __init__.py structure that seems to follow flask documentation. It is based on this post: https://realpython.com/blog/python/python-web-applications-with-flask-part-iii/

This structure should allow us to easily create a base test and inherit as they describe if we want to at some point later.

---

Review status: 2 of 8 files reviewed at latest revision, 6 unresolved discussions.

---

*[ufo/config.py, line 10 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x69L-xlavQfNUvAvM:-K80dC30APcnJCFTmJZP:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/ufo/config.py#L10)):*
Done.

---

*[ufo/config.py, line 16 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x6HtDYZDCyiWgyBxs:-K80dDwSEtHx0VgATcnL:536176272) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/ufo/config.py#L16)):*
Figured it out.

---

*[ufo/config.py, line 46 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x6G9NRPM02F4KgqLP:-K80dFMOQw9z9RSLnA0H:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/ufo/config.py#L46)):*
Done.

---

*[ufo/config.py, line 48 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x6ZiLXZsaxVexLMvW:-K80dH6LkGsfYbT3fhrI:536176272) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/ufo/config.py#L48)):*
Figured it out.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-:-K80cmtUKUinRcNXha1g:-187275819)*

jpevarnek:
FWIW, it's totally possible to do a flask app without code in __init__.py, that's just one way of setting it up.

https://github.com/sloria/cookiecutter-flask is a good example of a very well done skeleton that does not use it.

---

Reviewed 1 of 4 files at r2, 6 of 16 files at r4.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[.gitignore, line 6 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x5iFMPta8jehxkSX1:-K80ixkLHN2NyHxOqjFg:-1597320897) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/.gitignore#L6)):*
For the database, it might be easier to put anything in `instance` since that's already ignored and it is instance-specific.

For venv, can you remove that just since it should never be created and need to be ignored?  Thanks.

---

*[config.py, line 9 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80jgctS0a5ES5rGYla:-K80jgctS0a5ES5rGYlb:-635570777) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/config.py#L9)):*
Can you point this to the instance path?

---

*[config.py, line 11 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80jiqhgqA1pVYIgICy:-K80jiqhgqA1pVYIgICz:1107171968) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/config.py#L11)):*
This should be in version control, I don't think we need to keep it around as a comment.

---

*[config.py, line 29 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80joqRwv_7uzE-4xUP:-K80joqRwv_7uzE-4xUQ:246380807) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/config.py#L29)):*
I think this comment will be generall self-explanitory from the `BaseConfiguration` above, maybe just add a note on why we're overriding it for testing

---

*[config.py, line 31 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80jwlkvkXxFnoy9pRE:-K80jwlkvkXxFnoy9pRF:1324952482) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/config.py#L31)):*
nit: incredibly short lines here

---

*[ufo/__init__.py, line 23 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80jJjo8RbgqCGILxvi:-K80jJjo8RbgqCGILxvj:-823554315) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/ufo/__init__.py#L23)):*
Can you explain why these are template globals again?

---

*[ufo/user_test.py, line 35 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80kD0ccGBF0zV2kdBn:-K80kD0dTK3MP2TMFf87:843778248) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/ufo/user_test.py#L35)):*
Another case where I think us having the code in version control will be enough.

---

*[ufo/user_test.py, line 66 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80kOvK7iKW7vyXX4R-:-K80kOvK7iKW7vyXX4R0:-1935263732) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/ufo/user_test.py#L66)):*
Since we are expecting, for all n, that FAKE_EMAILS[n] and FAKE_NAMES[n] are linked, can we move these to a single object?

---

*[ufo/user_test.py, line 79 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80ki-tpRRdJrqmcfFm:-K80ki-tpRRdJrqmcfFn:1853039017) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/ufo/user_test.py#L79)):*
Is it possible to use `url_for` here instead of hardcoding a path (I can see the reasoning in wanting to test a hardcoded path, but I want to make sure this is either because we think it's the better way to do it or we think it's the only way)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-:-K80io7s4_VQ9rT9EALe:-659143008)*

ufo/user_test.py
```diff
@@ -0,0 +1,84 @@
...
+    db.session.delete(self.config)
+    db.session.commit()
+    db.session.close()
+
+  @patch.object(models.User, 'GetAll')
+  def testListUsersHandler(self, mock_get_all):
+    """Test the list user handler displays users from the database."""
+    fake_users = []
+    for x in range(0, len(FAKE_EMAILS)):
+      fake_user = MagicMock(id=x + 1, email=FAKE_EMAILS[x], name=FAKE_NAMES[x])
```
blueandgold: nit: a fake_user would be if you create a User() object.

for example
fake_user = User(fake_email, fake_name)

But because the use is a mock object, it's much more awesome to name it as mock_user
mock_user = MagicMock(id=x + 1, email=FAKE_EMAILS[x], name=FAKE_NAMES[x])

ufo/user_test.py
```diff
@@ -0,0 +1,84 @@
...
+      fake_users.append(fake_user)
```
blueandgold: nit: along the same line of thinking above, it would be more awesome to use "mock_users"

config.py
```diff
@@ -0,0 +1,34 @@
+from os.path import abspath, dirname, join
```
blueandgold: nit: It would be very useful to have here (or somehere else) to link in the awesome doc that describes why the config is structured like this.

ufo/user_test.py
```diff
@@ -0,0 +1,84 @@
...
+import os
+
+import flask
+from flask.ext.testing import TestCase
+from sqlalchemy import create_engine
+from sqlalchemy.orm import scoped_session
+from sqlalchemy.orm import sessionmaker
+import unittest
+
+from . import app, db
```
blueandgold: can we break up the import here into 2 lines?

from . import app
from . import db

xwsxethan:
Yeah I felt like I could have kept going with what I had and made it work, but the example I found seemed very straightforward, and I think it more closely resembles a lot of what I've found on SO already so it just made life easier. Plus it slimmed down this PR back to something more manageable.

---

---

*[.gitignore, line 6 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x5iFMPta8jehxkSX1:-K80lVOdKlns-PZZPdEC:-2098399017) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2aa1bafbb6e7c1a4d6efd40022dae4c9d5385b45/.gitignore#L6)):*
I'll remove each of these as the new tests run in memory so the db part should be unnecessary moving forward. Also, I found out .cache actually stores results from the latest test run. It will probably start appearing a directory up now that I modified Travis however, but it should be unnecessary moving forward. Done.

---

*[config.py, line 9 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80jgctS0a5ES5rGYla:-K80mLLs106eK22PCA3q:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/config.py#L9)):*
Done.

---

*[config.py, line 11 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80jiqhgqA1pVYIgICy:-K80mMd_aoxOIVAqO80z:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/config.py#L11)):*
Done.

---

*[config.py, line 29 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80joqRwv_7uzE-4xUP:-K80mkN4hHAY1vWXxq6F:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/config.py#L29)):*
Done.

---

*[config.py, line 31 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80jwlkvkXxFnoy9pRE:-K80mlMh0TthUnb-O678:1830810175) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/config.py#L31)):*
This was copy pasted from the site I mentioned in the comment. I adjusted it now. Done.

---

*[ufo/__init__.py, line 23 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80jJjo8RbgqCGILxvi:-K80mraA7NTycHOMvM8v:-2113850600) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/ufo/__init__.py#L23)):*
Template globals vs what exactly?

I put them in the template so they can be accessed conveniently in the html when we want to validate some kind of input. They can be used across a variety of pages so it seemed to make sense for them to be global rather than part of a specific response object or template. Is that what you're asking or do you mean something else?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-:-K80ndHmamzi-sBhbY43:728434231)*

blueandgold:
The testing setup looks so much cleaner than before.  Pending jonathan's comments, things look good on my end.  :+1:

xwsxethan:
Review status: 5 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[config.py, line 1 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80chTgZnklHy4d3TNf-r4-1:-K80pny-rRoeHkVZ6xvo:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/config.py#L1)):*
Done.

---

*[ufo/user_test.py, line 13 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x4hkawzIjmiUrV1Nf-r4-13:-K80ppzBjhO-_qYBpMX8:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/ufo/user_test.py#L13)):*
Done.

---

*[ufo/user_test.py, line 35 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80kD0ccGBF0zV2kdBn:-K80qFzOdKdI3Tm_NjPR:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/ufo/user_test.py#L35)):*
Done.

---

*[ufo/user_test.py, line 66 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x4hkawzIjmiUrV1Nf-r4-66:-K80pw0Z1UbZb339LK8B:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/ufo/user_test.py#L66)):*
Done.

---

*[ufo/user_test.py, line 66 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80kOvK7iKW7vyXX4R-:-K80r9EhZ8ixSw_HwtZz:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/ufo/user_test.py#L66)):*
Done.

---

*[ufo/user_test.py, line 67 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K7x4hkawzIjmiUrV1Nf-r4-67:-K80pyDkc2JI60bZGZ_V:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/ufo/user_test.py#L67)):*
Done.

---

*[ufo/user_test.py, line 79 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80ki-tpRRdJrqmcfFm:-K80rWCBcGUUzbajGViQ:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/ufo/user_test.py#L79)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4)*

jpevarnek:
One quick question, but :thumbsup: at your discretion.

---

Reviewed 1 of 3 files at r5, 1 of 1 files at r6.
Review status: 6 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[ufo/__init__.py, line 23 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80jJjo8RbgqCGILxvi:-K80vLSUhmiEXjpgeSzV:-359838505) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/ufo/__init__.py#L23)):*
As opposed to values within the specific templates.  At least for the textual error, I cannot imagine that being something that actually should be hard-coded between templates.  As far as the regular expressions, if we do end up doing that in enough different places, it seems like something that would be better suited to a Javascript library.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-:-K80vyguVmgzEywpgQuo:2143895269)*

jpevarnek:
Reviewed 1 of 3 files at r5.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4)*

xwsxethan:

---

*[ufo/__init__.py, line 23 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4#-K80jJjo8RbgqCGILxvi:-K80whAaMgjs1kXsG8MH:272113653) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/eb1726bf083412eb34011acbe1ce75dc327f3954/ufo/__init__.py#L23)):*
Ahh, I see what you're saying now. Yeah I'm fine with moving these over to javascript and then i18n-ing the errors once we figure out what exactly we're doing there. Most of the JS so far has been very small helper functions, so I was trying to avoid pushing more functionality over to that. I'll leave a TODO to clean that up once there's a clear path forward.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/4)*

PR #5, opened 2016-01-13
Switch to using helper methods on the database models

jpevarnek:
Since we are using a language that supports OO programming concepts, add
class methods to handle the operations that were in the database file

I left a few TODOs in here for places where I thought it would be much simpler to delete the `Model::GetById` method and switch to using `Model::query::get_or_404` in the actual methods.  Personally, I think it would be useful to get rid of both `Model::GetAll` and `Model::GetById` and then switch `Model::Add` to be `Model::save` and `Model::Delete` to be `Model::delete`.  I didn't want to change too much from what you had done, but would appreciate your input on those changes.


jpevarnek:
@xwsxethan would be great if you could also look at this

blueandgold:
:+1:  Fantastic, this does make the code look much nicer!  Good idea to change .Add() to .Save(), that is indeed better representation of what is happening.  As for the other GetAll() and GetById(), can we leave them in place and see how we can actually use them.

xwsxethan:
:+1: Looks good to me. Go ahead with merging this, and I'll adjust the test stuff in my PR as appropriate.

jpevarnek:
Henry, can you back this change out? I wanted to take care of a few more of
the review comments before moving it in. Thanks!

blueandgold:
Discussed with Jonathan.  The items are minor and will be simpler to address via a follow-up PR.

PR #7, opened 2016-01-14
Overload the database constructor (proxy serer in this case)

blueandgold:
This gives us an easier & cleaner way to create a db object from an request, as well as letting folks create the database object in whichever way they prefer.

``` python
# Way #1
server = models.ProxyServer.InitFromRequest(flask.request)

# Way #2
server = models.ProxyServer(
    name=flask.request.form.get('name'),
    ip_address=flask.request.form.get('ip_address'),
    ssh_private_key=flask.request.form.get('ssh_private_key'),
    fingerprint=flask.request.form.get('fingerprint'))

# Way #3
  server = models.ProxyServer()
  server.name=flask.request.form.get('name')
  server.ip_address=flask.request.form.get('ip_address')
  server.ssh_private_key=flask.request.form.get('ssh_private_key')
  server.fingerprint=flask.request.form.get('fingerprint')
```

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/7)


blueandgold:
@jpevarnek please take a look too

xwsxethan:
:+1:

jpevarnek:
Reviewed 2 of 3 files at r1, 1 of 1 files at r2.

---

*[.gitignore, line 3 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/7#-K80n-Nc7FyWoG-EQEmR:-K80n-Nc7FyWoG-EQEmS:-37861525) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/644433d160856d87ca9a31795105d44778f7199f/.gitignore#L3)):*
Future reference: this seems like it should be in a separate pull request.

---

*[ufo/models.py, line 111 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/7#-K80lZF947kjVVyeYwX_:-K80lZF947kjVVyeYwXa:-1222693390) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/644433d160856d87ca9a31795105d44778f7199f/ufo/models.py#L111)):*
I really don't think that this belongs in model code; this is very specific to the way we are structuring the form.

If we do want to do this, I think adding another function that extracts the values would be the best way to handle things.

In terms of the constructor, it might be easier to have something that takes kwargs and just passes those back to the base constructor if that's what we're worried about.  In the end, what we could end up with is (assuming you'd be fine with deleting both this and the `__init__` methods) is:

```
server = models.ProxyServer(flask.request.form)
server.save()
return flask.redirect(flask.url_for('proxyserver_list'))
```

I'd be in favor of having some validation of the request parameters beforehand, but that's all totally optional I guess.

---

*[ufo/proxy_server.py, line 51 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/7#-K80mvxtzXIaSHOQTa_A:-K80mvxtzXIaSHOQTa_B:-419358358) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/644433d160856d87ca9a31795105d44778f7199f/ufo/proxy_server.py#L51)):*
see above comment

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/7)*

goldspin:

---

*[.gitignore, line 3 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/7#-K80n-Nc7FyWoG-EQEmR:-K80r4m5ftzIKPw_9Zh-:-310385951) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/644433d160856d87ca9a31795105d44778f7199f/.gitignore#L3)):*
Acknowledged.

---

*[ufo/models.py, line 111 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/7#-K80lZF947kjVVyeYwX_:-K80rdegZnev920eOyAW:2128435507) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/644433d160856d87ca9a31795105d44778f7199f/ufo/models.py#L111)):*
Thanks for the review.  Yes, I can definitely pass the kwargs to a base constructor.  But from there, I'm not sure how to assign the kwargs to each different models, since each model has different attributes (i.e. the db columns).  Can you give me a bit more pointer on this?

For validation, I assumed we would be doing it on the client-side (i.e. with polymer).  What's the thought on doing validation on the server-side?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/7)*

jpevarnek:

---

*[ufo/models.py, line 111 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/7#-K80lZF947kjVVyeYwX_:-K80sx6LvAmVwy6sU6wR:-1705091183) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/644433d160856d87ca9a31795105d44778f7199f/ufo/models.py#L111)):*
I think client-side validation is a bit of a misnomer, it really should be called "client-side hinting" or something.  We cannot trust that any client-side validation has actually happened in the slightest and should take care of any verification on the server side.

As far as the different attributes, I'm a bit confused about what you mean.  Again, the above example will work perfectly fine with no changes other than deleting this function and the `__init__` functions from the code.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/7)*

blueandgold:
I'm going to defer this PR has it has validation implications that needs to be discussed.

PR #9, opened 2016-01-14
Ignore Eclipse Metafiles

blueandgold:


xwsxethan:
I'm not sure of the recursive syntax to match .project files in subdirectories, but if this works then go for it. :+1:

blueandgold:
Yes, this works in my eclipse env.  So, nothing recurive is needed.

PR #10, opened 2016-01-15
Start migrating the database tests to the flask version.

blueandgold:
Some small notes:
- @patch() seems to need a package reference, which wasn't necessary before.

``` python
@patch('ufo.models.RSA._RSAobj.exportKey')
```
- add back base64 encoding, which existed previously
- changed a couple of relative importing in the test seems not needed
- this is mainly for me to get a feel of the flask framework and the test setup

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/10)


blueandgold:
@jpevarnek PTAL too

jpevarnek:
Reviewed 2 of 2 files at r1.

---

*[ufo/models.py, line 70 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K81ztrwHRC1nUGDT7-S:-K81ztrwHRC1nUGDT7-T:-884856052) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models.py#L70)):*
Why are we base-64 encoding the data on disk?  To me, this seems pretty inefficient and it's better to store it in a binary format and encode it when that is necessary for display.

---

*[ufo/models_test.py, line 27 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K82-f6ixgPUtdwvpUx9:-K82-f6ixgPUtdwvpUxA:1584095905) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models_test.py#L27)):*
I think I saw this same function in tests from another PR, can we put this in a base class our test cases extend?

---

*[ufo/models_test.py, line 36 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K82-aBwJBkdas88fzK3:-K82-aBwJBkdas88fzK4:-1146967742) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models_test.py#L36)):*
whitespace

---

*[ufo/models_test.py, line 41 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K82-qp6rLe-AxgvIqLk:-K82-qp6rLe-AxgvIqLl:-1296849045) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models_test.py#L41)):*
Replacing this function seems to be incredibly relying upon an implementation detail, it seems like we can do a slightly better job of mocking this out.

I think what we are actually caring about for this test is that the value generated that is generated is a valid RSA key pair, I think we might actually be able to test that without any mocks.

I could be misunderstanding the point of this test though, it might help if you could explain that a bit further.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10)*

xwsxethan:
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[ufo/models_test.py, line 16 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K84tnda92h2WmIyEi-M:-K84tnda92h2WmIyEi-N:832939165) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models_test.py#L16)):*
I realize in looking back now that this is in user_test as well, but I don't think it is necessary, because of the switch to use config.py. You should be able to safely remove here, and I'll remove it in mine in a separate PR.

---

*[ufo/models_test.py, line 27 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K82-f6ixgPUtdwvpUx9:-K84sa6nxPYROt-BHfb3:-960057536) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models_test.py#L27)):*
The example I found showed a great way to do this with a base test class. It should be fairly straightforward. My one addition to this would be to move the code in user_test for setting up a config out to a base function that can be called when a child test needs that as part of setup. User_test (and other handler tests) would want that, whereas here we probably don't. Here's the example link for reference: https://realpython.com/blog/python/python-web-applications-with-flask-part-iii/

---

*[ufo/models_test.py, line 33 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K84uBXdDZlRNHWry1GN:-K84uBXdDZlRNHWry1GO:2068764868) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models_test.py#L33)):*
If you're only testing the db here, then this whole business with the config should not be necessary (unless one of the imports barfs without it). I needed this in user_test to get past the decorator, but that shouldn't be a problem here.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10)*

blueandgold:
Thanks for the review.  PTAL.  I posed a couple of questions back for discussion.  But definitely can be happy to go the other way.

---

Review status: 1 of 3 files reviewed at latest revision, 6 unresolved discussions.

---

*[ufo/models.py, line 70 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K81ztrwHRC1nUGDT7-S:-K85SCpQP2cUYT8GKDFH:-482097652) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models.py#L70)):*
The idea was to save the encoding step later on during request time, which seems to happen a lot more often.  Thoughts?  It's pretty easy to do either way, but want to double-check here.

---

*[ufo/models_test.py, line 16 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K84tnda92h2WmIyEi-M:-K85R_k1FoBs3S8Ixcki:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models_test.py#L16)):*
Done.

---

*[ufo/models_test.py, line 27 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K82-f6ixgPUtdwvpUx9:-K85UMfysVv68Ds-QKgk:2016723977) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models_test.py#L27)):*
Done, great idea.

---

*[ufo/models_test.py, line 33 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K84uBXdDZlRNHWry1GN:-K85UWAMTz-X-mtwVN8j:-961823460) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models_test.py#L33)):*
Done.  I moved the config stuff to a separate method, which you should be able to call in the handler tests.

---

*[ufo/models_test.py, line 36 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K82-aBwJBkdas88fzK3:-K85RfBFQcAYX_ZpoFfz:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models_test.py#L36)):*
Done.

---

*[ufo/models_test.py, line 41 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K82-qp6rLe-AxgvIqLk:-K85WyzWwElJB3JsAWqz:1977180856) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models_test.py#L41)):*
Correct, that we are testing our implementation only.  The assumption is that RSA library is working correctly and has been tested on their side.  i.e. it seems like of redundant of doing a functional testing such as sign a message with a key and then decode it with the other key.  We can do it this way, but we are just testing how that we are the library in the way that we desire, e.g. 2048 bit.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-:-K85Xi8JD0ajaJnMo8CC:1187603920)*

jpevarnek:
Reviewed 1 of 2 files at r2, 1 of 1 files at r3.

---

*[ufo/base_test.py, line 3 \[r3\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K85oL6SNWIroW_VWnz1:-K85oL6SNWIroW_VWnz2:-2060023780) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/be99d2283fc250cfd131148b92f3061778e60bf8/ufo/base_test.py#L3)):*
Can you remove the duplication from `user_test.py` with this?  Thanks.

---

*[ufo/models.py, line 70 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K81ztrwHRC1nUGDT7-S:-K85ny4wDUy_MKJ5hKSd:1616437224) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models.py#L70)):*
Sounds like a bit of premature optimization.

In general, let's try to keep the database layer concerning itself only with the actual data and not with the eventual representation we intend to use for that data in the UI.  That way, if we ever want to do something other than just showing the data to the user, we'll be able to without dealing with some strange format.

base-64 encoding should be essentially a no-op.

---

*[ufo/models_test.py, line 41 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K82-qp6rLe-AxgvIqLk:-K85optBrQt7-oteXu6F:-351610729) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models_test.py#L41)):*
I'm still having a bit of trouble understanding the purpose of this test.  It might help if you could clarify what kind of thing you'd expect this test to catch.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10)*

blueandgold:

---

*[ufo/models.py, line 70 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K81ztrwHRC1nUGDT7-S:-K86ED2fVfK9WH2x8bNP:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models.py#L70)):*
Done.

---

*[ufo/models_test.py, line 41 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K82-qp6rLe-AxgvIqLk:-K86EGXbq8up9HTZfCF9:1156447949) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models_test.py#L41)):*
The purpose of this test is that we are using the RSA library correctly and that we are returning object is in a format that we expect.

For example, as mentioned before, one of the assert is that we generating 2048 bit keys.

Another assert is that we are actually calling the library's apis correctly to get the right key pair.

The final assert is that the return object is what we expect.

As long as we test that this implementation is followed, we can accept that the keys are correct (without having to functionally verify that the library are returning valid keys, which we should trust is the case).

Thoughts?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10)*

blueandgold:
Review status: 1 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[ufo/base_test.py, line 3 \[r3\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K85oL6SNWIroW_VWnz1:-K86Ft-iofZSCXoLyG6l:1079334643) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/be99d2283fc250cfd131148b92f3061778e60bf8/ufo/base_test.py#L3)):*
Done.  I was intending to let Ethan do it when he updates the user_test.py, but guess I can now.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10)*

blueandgold:
Review status: 1 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[ufo/models_test.py, line 41 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K82-qp6rLe-AxgvIqLk:-K8QOIK4uRKhr2r89tlx:-600643934) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/c960eb66154812e8127c9ebbd7455ee2150a2baf/ufo/models_test.py#L41)):*
I have changed this test to a functional type (i.e. without mocking).

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10)*

blueandgold:
Also adding a testcase for regenerating the key pair.

xwsxethan:
Going ahead and :+1:  from me, but still pending Jonathan's comments.

---

Review status: 1 of 3 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-:-K8RHko7JX5rjIOKESn7:226701876)*

jpevarnek:
One quick comment, :thumbsup: at discretion

---

Reviewed 3 of 3 files at r4.

---

*[ufo/models_test.py, line 32 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K8RMJUdZNJRUtA5eYjA:-K8RMJUdZNJRUtA5eYjB:19064805) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/d7a4adc15e2ad6be8e04c217d65b9e73908f82e9/ufo/models_test.py#L32)):*
This looks like it is testing:

* The private implementation we have for generating a key pair, that should not be visible to tests, correctly returns an object that has these parameters as a valid key pair
* The RSA methods work

Honestly, I'm not sure how we could break this test in a way that would matter, I'd expect to be able to inline any private method and things to continue to work fine.  If the RSA library breaks...I think we just abandon hope at that point.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-:-K8RNCa85LG7zZZiQNwd:1609624754)*

goldspin:
Review status: 2 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[ufo/models_test.py, line 32 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10#-K8RMJUdZNJRUtA5eYjA:-K8RSDhHqfIvwO1IJE57:966844410) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/d7a4adc15e2ad6be8e04c217d65b9e73908f82e9/ufo/models_test.py#L32)):*
Done.  Not invoking the private method directly, and create a user object instead with the key pair present.

As for the other couple of details, I think that we should do some sanity testing that someone the returned key_pair doesn't have something like 2 public_keys, and that the key size is indeed 2048.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/10)*

PR #11, opened 2016-01-15
Migrate More User Tests Over to Flask

xwsxethan:
I'm going to add more to these in this same PR later, but I wanted to go ahead and get initial comments before I proceed too far so there's less work down the road.

Obviously, ignore all the commented out parts as those will migrated soon.

Please take a look at the change to xsrf and config as that was my easier way of disabling that during testing. We can similarly key off of any config value (it doesn't have to be WTF_CSRF_ENABLED for instance, it could be TESTING, or anything else), so long as it is different in production and testing modes.

Also note what I am asserting and what I am mocking. If this seems right then I'll continue as is, or if not, then I'll adjust what assertions we actually make moving forward. Thanks!

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/11)

### DIFF
--- a/ufo/user_test.py
+++ b/ufo/user_test.py
@@ -70,6 +70,217 @@ class UserTest(TestCase):
       details_link = flask.url_for('user_details', user_id=mock_users[x].id)
       self.assertEquals(details_link in user_list_output, True)
 
+  @patch.object(user, '_RenderUserAdd')
+  def testAddUsersGetHandler(self, mock_render):
+    """Test the add users get handler returns _RenderUserAdd's result."""
+    return_text = '<html>something here </html>'
+    mock_render.return_value = return_text
+    resp = self.client.get(flask.url_for('add_user'))
+
+    self.assertEquals(resp.data, return_text)
+
+  # @patch('user.User.InsertUsers')
+  # def testAddUsersPostHandler(self, mock_insert):
+  #   """Test the add users post handler calls to insert the specified users."""
+  #   user_1 = {}
+  #   user_1['primaryEmail'] = FAKE_EMAIL_1
+  #   user_1['name'] = {}
+  #   user_1['name']['fullName'] = FAKE_EMAIL_1
+  #   user_2 = {}
+  #   user_2['primaryEmail'] = FAKE_EMAIL_2
+  #   user_2['name'] = {}
+  #   user_2['name']['fullName'] = FAKE_EMAIL_2
+  #   user_array = []
+  #   user_array.append(user_1)
+  #   user_array.append(user_2)
+  #   data = '?selected_user={0}&selected_user={1}'.format(user_1, user_2)
+  #   response = self.testapp.post(PATHS['user_add_path'] + data)
+
+  #   mock_insert.assert_called_once_with(user_array)
+  #   self.assertEqual(response.status_int, 302)
+  #   self.assertTrue(PATHS['user_page_path'] in response.location)
+
+  # @patch('user.User.InsertUsers')
+  # def testAddUsersPostManualHandler(self, mock_insert):
+  #   """Test add users manually calls to insert the specified user."""
+  #   user_1 = {}
+  #   user_1['primaryEmail'] = FAKE_EMAIL
+  #   user_1['name'] = {}
+  #   user_1['name']['fullName'] = FAKE_NAME
+  #   user_array = []
+  #   user_array.append(user_1)
+  #   data = '?manual=true&user_name={0}&user_email={1}'.format(FAKE_NAME,
+  #                                                             FAKE_EMAIL)
+  #   response = self.testapp.post(PATHS['user_add_path'] + data)
+
+  #   mock_insert.assert_called_once_with(user_array)
+  #   self.assertEqual(response.status_int, 302)
+  #   self.assertTrue(PATHS['user_page_path'] in response.location)
+
+  # @patch('user._RenderAddUsersTemplate')
+  # @patch('google_directory_service.GoogleDirectoryService.WatchUsers')
+  # @patch('google_directory_service.GoogleDirectoryService.GetUserAsList')
+  # @patch('google_directory_service.GoogleDirectoryService.GetUsersByGroupKey')
+  # @patch('google_directory_service.GoogleDirectoryService.GetUsers')
+  # @patch('google_directory_service.GoogleDirectoryService.__init__')
+  # def testAddUsersGetHandlerNoParam(self, mock_ds, mock_get_users,
+  #                                   mock_get_by_key, mock_get_user,
+  #                                   mock_watch_users, mock_render):
+  #   """Test the add users get handler displays no users on initial get."""
+  #   # pylint: disable=too-many-arguments
+  #   mock_ds.return_value = None
+  #   self.testapp.get(PATHS['user_add_path'])
+
+  #   mock_ds.assert_called_once_with(MOCK_ADMIN.OAUTH_DECORATOR)
+  #   mock_get_users.assert_not_called()
+  #   mock_get_user.assert_not_called()
+  #   mock_get_by_key.assert_not_called()
+  #   mock_watch_users.assert_not_called()
+  #   mock_render.assert_called_once_with([])
+
+  # @patch('user._RenderAddUsersTemplate')
+  # @patch('google_directory_service.GoogleDirectoryService.WatchUsers')
+  # @patch('google_directory_service.GoogleDirectoryService.GetUserAsList')
+  # @patch('google_directory_service.GoogleDirectoryService.GetUsersByGroupKey')
+  # @patch('google_directory_service.GoogleDirectoryService.GetUsers')
+  # @patch('google_directory_service.GoogleDirectoryService.__init__')
+  # def testAddUsersGetHandlerWithGroup(self, mock_ds, mock_get_users,
+  #                                     mock_get_by_key, mock_get_user,
+  #                                     mock_watch_users, mock_render):
+  #   """Test the add users get handler displays users from a given group."""
+  #   # pylint: disable=too-many-arguments
+  #   mock_ds.return_value = None
+  #   # Email address could refer to group or user
+  #   group_key = 'foo@bar.mybusiness.com'
+  #   mock_get_by_key.return_value = FAKE_USER_ARRAY
+  #   self.testapp.get(PATHS['user_add_path'] + '?group_key=' + group_key)
+
+  #   mock_get_users.assert_not_called()
+  #   mock_get_user.assert_not_called()
+  #   mock_ds.assert_called_once_with(MOCK_ADMIN.OAUTH_DECORATOR)
+  #   mock_get_by_key.assert_called_once_with(group_key)
+  #   mock_watch_users.assert_any_call('delete')
+  #   mock_watch_users.assert_any_call('makeAdmin')
+  #   mock_watch_users.assert_any_call('undelete')
+  #   mock_watch_users.assert_any_call('update')
+  #   mock_render.assert_called_once_with(FAKE_USER_ARRAY)
+
+  # @patch('user._RenderAddUsersTemplate')
+  # @patch('google_directory_service.GoogleDirectoryService.WatchUsers')
+  # @patch('google_directory_service.GoogleDirectoryService.GetUserAsList')
+  # @patch('google_directory_service.GoogleDirectoryService.GetUsersByGroupKey')
+  # @patch('google_directory_service.GoogleDirectoryService.GetUsers')
+  # @patch('google_directory_service.GoogleDirectoryService.__init__')
+  # def testAddUsersGetHandlerWithUser(self, mock_ds, mock_get_users,
+  #                                    mock_get_by_key, mock_get_user,
+  #                                    mock_watch_users, mock_render):
+  #   """Test the add users get handler displays a given user as requested."""
+  #   # pylint: disable=too-many-arguments
+  #   mock_ds.return_value = None
+  #   # Email address could refer to group or user
+  #   user_key = 'foo@bar.mybusiness.com'
+  #   mock_get_user.return_value = FAKE_USER_ARRAY
+  #   self.testapp.get(PATHS['user_add_path'] + '?user_key=' + user_key)
+
+  #   mock_get_users.assert_not_called()
+  #   mock_get_by_key.assert_not_called()
+  #   mock_ds.assert_called_once_with(MOCK_ADMIN.OAUTH_DECORATOR)
+  #   mock_get_user.assert_called_once_with(user_key)
+  #   mock_watch_users.assert_any_call('delete')
+  #   mock_watch_users.assert_any_call('makeAdmin')
+  #   mock_watch_users.assert_any_call('undelete')
+  #   mock_watch_users.assert_any_call('update')
+  #   mock_render.assert_called_once_with(FAKE_USER_ARRAY)
+
+  # @patch('user._RenderAddUsersTemplate')
+  # @patch('google_directory_service.GoogleDirectoryService.WatchUsers')
+  # @patch('google_directory_service.GoogleDirectoryService.GetUserAsList')
+  # @patch('google_directory_service.GoogleDirectoryService.GetUsersByGroupKey')
+  # @patch('google_directory_service.GoogleDirectoryService.GetUsers')
+  # @patch('google_directory_service.GoogleDirectoryService.__init__')
+  # def testAddUsersGetHandlerWithAll(self, mock_ds, mock_get_users,
+  #                                   mock_get_by_key, mock_get_user,
+  #                                   mock_watch_users, mock_render):
+  #   """Test the add users get handler displays all users in a domain."""
+  #   # pylint: disable=too-many-arguments
+  #   mock_ds.return_value = None
+  #   mock_get_users.return_value = FAKE_USER_ARRAY
+  #   self.testapp.get(PATHS['user_add_path'] + '?get_all=true')
+
+  #   mock_get_by_key.assert_not_called()
+  #   mock_get_user.assert_not_called()
+  #   mock_ds.assert_called_once_with(MOCK_ADMIN.OAUTH_DECORATOR)
+  #   mock_get_users.assert_called_once_with()
+  #   mock_watch_users.assert_any_call('delete')
+  #   mock_watch_users.assert_any_call('makeAdmin')
+  #   mock_watch_users.assert_any_call('undelete')
+  #   mock_watch_users.assert_any_call('update')
+  #   mock_render.assert_called_once_with(FAKE_USER_ARRAY)
+
+  # @patch('user._RenderAddUsersTemplate')
+  # @patch('google_directory_service.GoogleDirectoryService.WatchUsers')
+  # @patch('google_directory_service.GoogleDirectoryService.GetUserAsList')
+  # @patch('google_directory_service.GoogleDirectoryService.GetUsersByGroupKey')
+  # @patch('google_directory_service.GoogleDirectoryService.GetUsers')
+  # @patch('google_directory_service.GoogleDirectoryService.__init__')
+  # def testAddUsersGetHandlerWithError(self, mock_ds, mock_get_users,
+  #                                     mock_get_by_key, mock_get_user,
+  #                                     mock_watch_users, mock_render):
+  #   """Test the add users get handler fails gracefully."""
+  #   # pylint: disable=too-many-arguments
+  #   fake_status = '404'
+  #   fake_response = MagicMock(status=fake_status)
+  #   fake_content = b'some error content'
+  #   fake_error = errors.HttpError(fake_response, fake_content)
+  #   mock_ds.side_effect = fake_error
+  #   mock_get_users.return_value = FAKE_USER_ARRAY
+  #   self.testapp.get(PATHS['user_add_path'] + '?get_all=true')
+
+  #   mock_ds.assert_called_once_with(MOCK_ADMIN.OAUTH_DECORATOR)
+  #   mock_get_by_key.assert_not_called()
+  #   mock_get_user.assert_not_called()
+  #   mock_get_users.assert_not_called()
+  #   mock_watch_users.assert_not_called()
+  #   mock_render.assert_called_once_with([], fake_error)
+
+  # @patch('user.User.InsertUsers')
+  # def testAddUsersPostHandler(self, mock_insert):
+  #   """Test the add users post handler calls to insert the specified users."""
+  #   user_1 = {}
+  #   user_1['primaryEmail'] = FAKE_EMAIL_1
+  #   user_1['name'] = {}
+  #   user_1['name']['fullName'] = FAKE_EMAIL_1
+  #   user_2 = {}
+  #   user_2['primaryEmail'] = FAKE_EMAIL_2
+  #   user_2['name'] = {}
+  #   user_2['name']['fullName'] = FAKE_EMAIL_2
+  #   user_array = []
+  #   user_array.append(user_1)
+  #   user_array.append(user_2)
+  #   data = '?selected_user={0}&selected_user={1}'.format(user_1, user_2)
+  #   response = self.testapp.post(PATHS['user_add_path'] + data)
+
+  #   mock_insert.assert_called_once_with(user_array)
+  #   self.assertEqual(response.status_int, 302)
+  #   self.assertTrue(PATHS['user_page_path'] in response.location)
+
+  # @patch('user.User.InsertUsers')
+  # def testAddUsersPostManualHandler(self, mock_insert):
+  #   """Test add users manually calls to insert the specified user."""
+  #   user_1 = {}
+  #   user_1['primaryEmail'] = FAKE_EMAIL
+  #   user_1['name'] = {}
+  #   user_1['name']['fullName'] = FAKE_NAME
+  #   user_array = []
+  #   user_array.append(user_1)
+  #   data = '?manual=true&user_name={0}&user_email={1}'.format(FAKE_NAME,
+  #                                                             FAKE_EMAIL)
+  #   response = self.testapp.post(PATHS['user_add_path'] + data)
+
+  #   mock_insert.assert_called_once_with(user_array)
+  #   self.assertEqual(response.status_int, 302)
+  #   self.assertTrue(PATHS['user_page_path'] in response.location)
+
 
 if __name__ == '__main__':
   unittest.main()
--- a/ufo/user_test.py
+++ b/ufo/user_test.py
@@ -5,6 +5,7 @@ import os
 
 import flask
 from flask.ext.testing import TestCase
+from googleapiclient import errors
 import json
 from sqlalchemy import create_engine
 from sqlalchemy.orm import scoped_session
@@ -40,7 +41,7 @@ for x in range(0, len(FAKE_EMAILS_AND_NAMES)):
   fake_add_user['type'] = 'USER'
   FAKE_DIRECTORY_USER_ARRAY.append(fake_add_user)
 
-FAKE_CREDENTIALS = 'Look at me, Im a credential'
+FAKE_CREDENTIALS = 'Look at me. I am a credential!'
 
 class UserTest(TestCase):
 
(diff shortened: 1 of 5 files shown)


xwsxethan:
@jpevarnek PTAL as well.

Please both of you check my comment on the PR as this is not all the code I wanted to merge over, but just wanted to get initial feedback. Thanks!

jpevarnek:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[config.py, line 14 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85pFaEY0RjMQWMNCXZ:-K85pFaFroSAWYYzBE_1:486893205) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/config.py#L14)):*
`TESTING` seems to be used in a couple other modules, might be useful to just use the same thing here.

---

*[ufo/user_test.py, line 43 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85pjWWK434i_oy719G:-K85pjWWK434i_oy719H:322494464) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L43)):*
nit: punctuation

---

*[ufo/user_test.py, line 72 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85pxSADEtMutpnsy5M:-K85pxSADEtMutpnsy5N:492111629) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L72)):*
`s/displays/should display/`

---

*[ufo/user_test.py, line 74 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85q00C2LLocDxi-vT7:-K85q00DFY32JnY0I039:-21791096) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L74)):*
Can you do this off of the public method?

Also, might be useful to turn off template rendering and check the parameters there in some way.  This test could change incredibly with internationalization or even just with us adding a line break somewhere.

---

*[ufo/user_test.py, line 115 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85r9ubiAuRLnz4GHbm:-K85r9ubiAuRLnz4GHbn:-1741852201) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L115)):*
Same comment as above regarding `_RenderUserAdd`

---

*[ufo/user_test.py, line 119 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85quueHZvNgHkHjf3S:-K85quueHZvNgHkHjf3T:-1633400647) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L119)):*
My personal take on this is that we don't actually want any of the above lines, they're all implementation details we should be able to change.  As far as what we care about, I think the only thing that actually matters is we should be outputting whatever we get from the get by group key function

---

*[ufo/user_test.py, line 226 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85rUzb-ya77THOrGhk:-K85rUzb-ya77THOrGhl:776386639) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L226)):*
Again, I have a tiny issue with us knowing the `_RenderUserAdd` function exists.  The fact that I could break all the tests by inlining a no-public function terrifies me.

---

*[ufo/user_test.py, line 255 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85ruHpFKHi48qfHA4n:-K85ruHpFKHi48qfHA4o:2124105984) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L255)):*
My personal instinct with the test condition here is that we should be checking either that we are then able to get information about the user or that it now exists in the database with the expected properties.  If that does not feel right to you, I'm fine sticking with the way this is

---

*[ufo/xsrf.py, line 13 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85sblZVdnBi0hl1OY9:-K85sblZVdnBi0hl1OYA:189505174) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/xsrf.py#L13)):*
To get  rid of some indentation creep:
```
if app.config['TESTING']:
  return
```

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11)*

jpevarnek:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[ufo/xsrf.py, line 13 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85sblZVdnBi0hl1OY9:-K85sqKcAHyJkSIhRo6N:-162456924) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/xsrf.py#L13)):*
(or
```
if not app.config['WTF_CSRF_ENABLED']:
  return
```
)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11)*

xwsxethan:

---

*[config.py, line 14 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85pFaEY0RjMQWMNCXZ:-K85vAEGRl4fUeY1WyA6:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/config.py#L14)):*
Done.

---

*[ufo/user_test.py, line 43 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85pjWWK434i_oy719G:-K85vEMQtUOuVtVbkKba:-1585674380) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L43)):*
Lol, done.

---

*[ufo/user_test.py, line 72 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85pxSADEtMutpnsy5M:-K85vcBHgPrHgNiaxbMB:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L72)):*
Done.

---

*[ufo/user_test.py, line 74 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85q00C2LLocDxi-vT7:-K85vkh9GCLux36rmVvl:-955800380) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L74)):*
As for going off of the public method, yeah probably. Doing it this way just felt like each test could be a bit smaller and require less mocking, but I understand the reasoning for not wanting to directly call a private function. I'll switch this and the others over to that as per the comments.

For internationalization, my thought is that we may be able to change the code on the backend entirely without doing anything client side. Henry had some ideas for this that he had used before. So that being said, my hope is that this would progress to me referencing a constant whose value is 'OAuth is not set up' instead and asserting that the constant is present in the rendered html (similarly, if the constant references a placeholder which points to the i18n value, I just assert that the placeholder is present or translated in both).

Would that seem like a reasonable approach? Otherwise, my only assertion here would be that render was called with nothing or that no users in the datastore were present, which is a little tricky.

---

*[ufo/user_test.py, line 119 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85quueHZvNgHkHjf3S:-K85ws40N-PfpYJOaiMI:926664104) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L119)):*
That's how I felt about the assertion on mock ds, but I wasn't sure if asserting that the other methods on the directory service weren't called felt valid. If I'm reading the tests like a spec for what this service does, then without those asserts, it isn't directly clear that only call the matching one (vs calling all of them on every request and picking some result). That detail may not be crucial however, since this is somewhat implied by only mock_get_by_key having a specified return value.

---

*[ufo/user_test.py, line 255 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85ruHpFKHi48qfHA4n:-K85uKbzy1N_Rk7uWNfk:-1891524232) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L255)):*
I felt that way a little bit while doing the test, but my instinct says that feels more like a functional test since I'm then depending on the db behaving correctly, which should itself be tested elsewhere. This felt like the only thing I could assert on, since I don't really care much about how the data gets parsed and formatted, just that it goes there.

---

*[ufo/xsrf.py, line 13 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85sblZVdnBi0hl1OY9:-K85v4t34qRyd5vqFdsO:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/xsrf.py#L13)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11)*

jpevarnek:
I think you forgot to push.

---

---

*[ufo/user_test.py, line 74 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85q00C2LLocDxi-vT7:-K8605JR9ENFGrS7X-JH:398257461) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L74)):*
My way of checking it would have just been to ensure that the `directory_users` parameter is an empty array and that `error` is set, does that sounds reasonable?

---

*[ufo/user_test.py, line 119 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85quueHZvNgHkHjf3S:-K86-oXxG29l65C3g7VW:847441104) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L119)):*
If you want to test that we're not making unnecessary calls, I'd split that into a separate test than the one that checks that we are outputting the correct information.

---

*[ufo/user_test.py, line 255 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85ruHpFKHi48qfHA4n:-K86-nvFBdjQWq7g6qLj:-1688847758) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L255)):*
I think it's fine assuming that the SQLAlchemy is working correctly given it has its own tests.

I think we'd be pretty fine following the example from http://pythonhosted.org/Flask-Testing/#testing-with-sqlalchemy

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-:-K860T7XmeyRAIltCceb:-1151593631)*

blueandgold:

---

*[ufo/user_test.py, line 227 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K86pEybrPSJWDChHyU8:-K86pEycvqT8o6mNKTHw:-825724413) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L227)):*
This is really cool.  Yes, how the one handler determines a get or post absoutely needs to have test coverage.

---

*[ufo/user_test.py, line 255 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85ruHpFKHi48qfHA4n:-K86gGpxKjI4W_UDUrTB:1383240641) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L255)):*
This all depends how we structure things.  :)  Testing the handlers with a real database is not a bad idea if we can wire up everything the test needs, and the test would become a functional test.  This seems to stretch the definition of unit testing.  For example, in this case, the handler would save our users, do the redirect, and then we check the list page has the added users.

Also, the handler code here seems to have changed.  Previously, there was an InsertUsers() method, which abstracted the user instance creation on the datastore(base) side.

Now, the handler code seems to be iterating and adding each user one at at time to the database (i.e. via multiple commits).

So, this is something else we can discuss next week!  Whether this is the right way or there's something better.  In general, the handlers here represent a frontend client, and there should be as little logic here as possible.  In my perspective, there can be other clients, e.g. mobile android, ios, etc, or even a CLI tool.  We wouldn't want each client to implement their own database save and/or validation logic, etc.  In this approach, there should be a backend where the application logic lives and provides api to the frontend client, and then calls the actual storage.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11)*

xwsxethan:
Sorry for the confusion. I wasn't going to push until I implemented more along your guidelines, but I was just marking things done as I took care of them in the code. You should be able to see those now.

---

Review status: 1 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[ufo/user_test.py, line 74 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85q00C2LLocDxi-vT7:-K8R0GmZcfFoEkf-TN4J:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L74)):*
Done.

---

*[ufo/user_test.py, line 119 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85quueHZvNgHkHjf3S:-K8R0JO55j2SB2C7cgIC:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L119)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-:-K8R0Sl1Tc0XBE12-GqK:2082753247)*

ufo/user_test.py
```diff
@@ -22,6 +30,18 @@
...
+FAKE_DIRECTORY_USER_ARRAY = []
+for x in range(0, len(FAKE_EMAILS_AND_NAMES)):
+  fake_add_user = {}
+  fake_add_user['primaryEmail'] = FAKE_EMAILS_AND_NAMES[x]['email']
+  fake_add_user['name'] = {}
+  fake_add_user['name']['fullName'] = FAKE_EMAILS_AND_NAMES[x]['name']
+  fake_add_user['email'] = FAKE_EMAILS_AND_NAMES[x]['email']
+  fake_add_user['role'] = 'MEMBER'
+  fake_add_user['type'] = 'USER'
+  FAKE_DIRECTORY_USER_ARRAY.append(fake_add_user)
```
blueandgold: nit: this line would read much better if "fake_add_user" is "fake_directory_user"
FAKE_DIRECTORY_USER_ARRAY.append(fake_directory_user)

ufo/user_test.py
```diff
@@ -70,6 +90,167 @@ def testListUsersHandler(self, mock_get_all):
...
+    """Test add user get should display an error when oauth isn't set."""
+    mock_get_saved_credentials.return_value = None
+    mock_render_template.return_value = ''
+
+    response = self.client.get(flask.url_for('add_user'))
+
+    args, kwargs = mock_render_template.call_args
+    self.assertEquals('add_user.html', args[0])
+    self.assertEquals([], kwargs['directory_users'])
+    self.assertEquals(kwargs['error'] is not None, True)
```
blueandgold: nit: assertIsNotNone
https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertIsNotNone

self.assertIsNotNone(kwargs['error'])

ufo/user_test.py
```diff
@@ -70,6 +90,167 @@ def testListUsersHandler(self, mock_get_all):
...
+
+  @patch('flask.render_template')
+  @patch.object(oauth, 'getSavedCredentials')
+  @patch.object(gds.GoogleDirectoryService, '__init__')
+  def testAddUsersGetNoParam(self, mock_ds, mock_get_saved_credentials,
+                            mock_render_template):
+    """Test add user get should display no users on initial get."""
+    mock_get_saved_credentials.return_value = FAKE_CREDENTIALS
```
blueandgold: nit: plural?  Because when it's plural, I would expect a list.  But it's defined as a singleton string.

ufo/user_test.py
```diff
@@ -70,6 +90,167 @@ def testListUsersHandler(self, mock_get_all):
       details_link = flask.url_for('user_details', user_id=mock_users[x].id)
       self.assertEquals(details_link in user_list_output, True)
 
+  @patch.object(user, '_RenderUserAdd')
+  def testAddUsersGetHandler(self, mock_render):
+    """Test the add users get handler returns _RenderUserAdd's result."""
+    return_text = '<html>something here </html>'
+    mock_render.return_value = return_text
```
blueandgold: nit: If I understand Jonathan's thoughts correctly, this should render a real template instead of a mock?  But let's see what he thinks. 
@jpevarnek

xwsxethan:
Review status: 1 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[ufo/user_test.py, line 42 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85kOEaOXzJMmS008ds-r2-42:-K8RGXiQsD1Tl_K0Gojk:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/bf057009a88c5314a7d9c4db71e276032a20d903/ufo/user_test.py#L42)):*
Done.

---

*[ufo/user_test.py, line 115 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85kOEaOXzJMmS008ds-r2-115:-K8RGZQx5L7ZDciw1KxR:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/bf057009a88c5314a7d9c4db71e276032a20d903/ufo/user_test.py#L115)):*
Done.

---

*[ufo/user_test.py, line 123 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85kOEaOXzJMmS008ds-r2-123:-K8RG_l6zl8mACwclyTy:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/bf057009a88c5314a7d9c4db71e276032a20d903/ufo/user_test.py#L123)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11)*

blueandgold:
:+1: Things look good by me.  But please have Jonathan to have a look as well.

jpevarnek:
Clarification on the template rendering: I was thinking closer to http://flask-testing.readthedocs.org/en/latest/#opt-to-not-render-the-templates, your call if you think there is a reason this will work better.

I just skimmed some of the parts since my flight is about to take off...I'll try looking at any pain points more once we're in the air :)

---

Reviewed 1 of 3 files at r2.
Review status: 2 of 4 files reviewed at latest revision, 8 unresolved discussions.

---

*[config.py, line 15 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8RD2BwwtIhr4owxfcl:-K8RD2BwwtIhr4owxfcm:1596519549) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/bf057009a88c5314a7d9c4db71e276032a20d903/config.py#L15)):*
Do we still need the `WTF...` variable?

---

*[ufo/user_test.py, line 255 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85ruHpFKHi48qfHA4n:-K8RKDitzOoPNoOm0Z10:-1028847118) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L255)):*
Definitely shouldn't need to be something that has multiple commits, it would actually be much simpler to just queue those all up and commit at the end.  But yes, let's talk about this tomorrow because I think there's a bit of confusion about what we consider to be the logic in the API.

---

*[ufo/user_test.py, line 97 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85kOEaOXzJMmS008ds-r2-97:-K8RHi07YXEoNCLNfweE:1795904617) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/bf057009a88c5314a7d9c4db71e276032a20d903/ufo/user_test.py#L97)):*
My comment was more that we should not be paying attention to the private implementation methods.  My reasoning:

What this test actually does is test that the output of some private method we should not know about is the same as the output of a URL handler.

If I choose to inline the code, that will completely and totally break everything.  In my opinion, changing a private implementation detail should not break tests.

If I delete the code that actually renders the template from `_RenderUserAdd`, that will have actually no effect on this template.  Deleting a major piece of functionality is something I am totally fine with breaking tests (and would hope it would).  At the point where that basic functionality has been tested, then I'd say it's fine to mock it out.

---

*[ufo/user_test.py, line 85 \[r3\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8RH8mEpQtJjFQ_h3hD:-K8RH8mEpQtJjFQ_h3hE:1828453666) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a3aed67455bf626f29c8c7c7d86383b27dcd24d0/ufo/user_test.py#L85)):*
The `x + 1` reads really oddly here with the spacing...maybe put parentheses around it?

---

*[ufo/user_test.py, line 112 \[r3\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8RId7GUOkchVi_k6xe:-K8RId7GUOkchVi_k6xf:-1104293015) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a3aed67455bf626f29c8c7c7d86383b27dcd24d0/ufo/user_test.py#L112)):*
For this and a few tests below it, see main comment above.

---

*[ufo/user_test.py, line 159 \[r3\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8RIjnKcQOKy6ASpoc9:-K8RIjnKcQOKy6ASpocA:-749317342) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a3aed67455bf626f29c8c7c7d86383b27dcd24d0/ufo/user_test.py#L159)):*
`self.client.get(flask.url_for('add_user', group_key=group_key))`

---

*[ufo/user_test.py, line 181 \[r3\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8RJoRKPla1KDM7Iar2:-K8RJoRKPla1KDM7Iar3:2107608502) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a3aed67455bf626f29c8c7c7d86383b27dcd24d0/ufo/user_test.py#L181)):*
Pass arguments to `url_for`

---

*[ufo/user_test.py, line 201 \[r3\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8RJvsdlJo0Wm4PqZvK:-K8RJvsdlJo0Wm4PqZvL:-469765787) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a3aed67455bf626f29c8c7c7d86383b27dcd24d0/ufo/user_test.py#L201)):*
ditto

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-:-K8RDoJFKgXiQp-Hytm6:-1379131987)*

xwsxethan:
Review status: 2 of 4 files reviewed at latest revision, 6 unresolved discussions.

---

*[config.py, line 15 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8RD2BwwtIhr4owxfcl:-K8RMDhApzK3RFA7Xiid:-982084450) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/bf057009a88c5314a7d9c4db71e276032a20d903/config.py#L15)):*
No, I removed it now as a result. Done.

---

*[ufo/user_test.py, line 85 \[r3\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8RH8mEpQtJjFQ_h3hD:-K8RMX8WcLJGI_bJ7xxR:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a3aed67455bf626f29c8c7c7d86383b27dcd24d0/ufo/user_test.py#L85)):*
Done.

---

*[ufo/user_test.py, line 159 \[r3\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8RIjnKcQOKy6ASpoc9:-K8RNZtGDlNoV6T-XjwG:-1151641289) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a3aed67455bf626f29c8c7c7d86383b27dcd24d0/ufo/user_test.py#L159)):*
This actually doesn't work. When building the environment for the request, it throws errors. I added the variable into the add_user function definition, but I don't think it will work without making it part of the route, like /user/add/<group_key> or /user/add/<user_key> which seems weird with so many possibilities. Are we fine leaving these as just query strings instead?

---

*[ufo/user_test.py, line 181 \[r3\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8RJoRKPla1KDM7Iar2:-K8RRKd-tPw4qwyqz0lE:-1582418045) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a3aed67455bf626f29c8c7c7d86383b27dcd24d0/ufo/user_test.py#L181)):*
See above.

---

*[ufo/user_test.py, line 201 \[r3\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8RJvsdlJo0Wm4PqZvK:-K8RRLj9NFAdoD8R02DS:-1582418045) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a3aed67455bf626f29c8c7c7d86383b27dcd24d0/ufo/user_test.py#L201)):*
See above.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11)*

jpevarnek:
Okay, looks good.  Still confused at why that's breaking, I'll try to take a look at the errors there locally.

---

Reviewed 1 of 2 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[ufo/user_test.py, line 159 \[r3\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8RIjnKcQOKy6ASpoc9:-K8RWDL6-un-oset38x5:380666228) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a3aed67455bf626f29c8c7c7d86383b27dcd24d0/ufo/user_test.py#L159)):*
Hmm, oddly, this seems to be working fine locally...

What exactly are the errors you are seeing?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-:-K8RXOFPl3yTFuosJOje:112750656)*

xwsxethan:
Review status: 3 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[ufo/user_test.py, line 159 \[r3\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8RIjnKcQOKy6ASpoc9:-K8UTOgaYnbUgiNJqDhL:1042778433) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a3aed67455bf626f29c8c7c7d86383b27dcd24d0/ufo/user_test.py#L159)):*
Addressed as per the comments here: https://github.com/uProxy/ufo-management-server-flask/pull/13

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11)*

jpevarnek:
I just glanced through the whole thing again.  Couple questions, but I think everything looks good now!

---

Reviewed 3 of 3 files at r7.
Review status: 2 of 5 files reviewed at latest revision, 9 unresolved discussions.

---

*[ufo/templates/user_details.html, line 9 \[r7\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8UoCuLBS6Qbj87uWs8:-K8UoCuLBS6Qbj87uWs9:-1855249180) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9d65fa2f0844eaa7f83453c7ecbd2267f21f4897/ufo/templates/user_details.html#L9)):*
Out of curiosity, why are we doing this instead of just calling `xsrf_token` below?

---

*[ufo/user.py, line 144 \[r7\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8Up13ZAPf17ZufqlZV:-K8Up13__AXv1PH1-qho:863154377) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9d65fa2f0844eaa7f83453c7ecbd2267f21f4897/ufo/user.py#L144)):*
:heart:

---

*[ufo/user_test.py, line 19 \[r7\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8UpBwVqtkMNCBunVsM:-K8UpBwVqtkMNCBunVsN:1548535385) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9d65fa2f0844eaa7f83453c7ecbd2267f21f4897/ufo/user_test.py#L19)):*
Why does this get imported from `.`? (this is just me not understanding Python...)

---

*[ufo/user_test.py, line 50 \[r7\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8UpfXlGwtW5jr_VBww:-K8UpfXlGwtW5jr_VBwx:1241201498) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9d65fa2f0844eaa7f83453c7ecbd2267f21f4897/ufo/user_test.py#L50)):*
This seems a bit odd to have as a global variable, what's the reasoning for that?

---

*[ufo/user_test.py, line 110 \[r7\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8UqGN7WV7UiSv_AP7q:-K8UqGN8dVY3vlQBeFs5:1930201802) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9d65fa2f0844eaa7f83453c7ecbd2267f21f4897/ufo/user_test.py#L110)):*
I feel like I normally see the expected value as the second parameter in test asserts, does Python handle printing errors correctly with it as the first one (normally I'll see errors like "Expected {first} to be {second}")?

---

*[ufo/user_test.py, line 196 \[r7\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8UqqliWeOkbg19jtaU:-K8UqqliWeOkbg19jtaV:-1064181891) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9d65fa2f0844eaa7f83453c7ecbd2267f21f4897/ufo/user_test.py#L196)):*
Can you be a bit more specific for what kind of error this is handling?  Honestly, it took a few read-throughts to figure out that it's looking for the resource not being found by directory service (also, would https://developers.google.com/api-client-library/python/guide/mocks be useful or just more effort?)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-:-K8UoHlRLO2unjpoJ9m7:-1575261547)*

xwsxethan:
Review status: 2 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[ufo/templates/user_details.html, line 9 \[r7\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8UoCuLBS6Qbj87uWs8:-K8UvLNsV4Qv-LhMrm3_:1132328865) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9d65fa2f0844eaa7f83453c7ecbd2267f21f4897/ufo/templates/user_details.html#L9)):*
This is so the token gets defined once per page and doesn't invalidate all but the last form.

I discovered when playing with the add_user page that calling xsrf_token for each form causes multiple tokens to be generated which effectively renders all except for the most recently generated one to be invalid. So depending on how the template is loaded/executed various posts will fail with invalid xsrf tokens since they are not the newest one, which is expected.

Similarly since there are multiple forms on this page now (for toggle revoked, delete user, and get new key pair), I have to set the token once and then supply that token in whatever form gets submitted.

I could probably move this functionality to the base page since it seems like we will have to use it in multiple places eventually, but for right now this seems just as easy.

---

*[ufo/user_test.py, line 19 \[r7\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8UpBwVqtkMNCBunVsM:-K8Uya2T9SYBvVXNnR6a:-1086231363) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9d65fa2f0844eaa7f83453c7ecbd2267f21f4897/ufo/user_test.py#L19)):*
I'll be honest, I don't really understand python imports entirely myself. It seems that importing from . or just importing directly works for other files. For app and db which are in __init__.py, they have to be reference from . however. I changed the others over though since that is more clear.

---

*[ufo/user_test.py, line 50 \[r7\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8UpfXlGwtW5jr_VBww:-K8UwS_K88dcEeyMW9GB:-855147954) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9d65fa2f0844eaa7f83453c7ecbd2267f21f4897/ufo/user_test.py#L50)):*
It was just so I could reference the value later in the tests if needed, though I don't think I ever did. I just supplied them directly to the mock now.

---

*[ufo/user_test.py, line 110 \[r7\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8UqGN7WV7UiSv_AP7q:-K8Uwr0HddTOcOg0nWGk:288857273) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9d65fa2f0844eaa7f83453c7ecbd2267f21f4897/ufo/user_test.py#L110)):*
Hmmm, for the way we're running the tests and asserts, python doesn't seem to print out in that format. It just says 'AssertionError: foo != bar'. We can change them to have the expectation second, but for printing purposes it won't really read any better since it doesn't specify which is the expectation vs the test value. Your call on what to do here.

As a side note, it does seem to print more like an expectation when other things fail. For instance, if we call assert_called_once_with, and the method is actually called twice, it will print "AssertionError: Expected 'foo' to be called once. Called 2 times." So some errors are printed more legibly, just not assertEquals.

---

*[ufo/user_test.py, line 196 \[r7\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8UqqliWeOkbg19jtaU:-K8UzYSFCvvadCbF3BfO:975666750) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9d65fa2f0844eaa7f83453c7ecbd2267f21f4897/ufo/user_test.py#L196)):*
I'll expand the docstring to explain this more clearly. Yeah so because we didn't have robust error handling yet, when a resource that we searched for in the directory service wasn't found, we were bubbling that up as a 404 rather than saying 'user/group foo not found. search again.'. I implemented logic to catch that error and still return the page gracefully instead of looking like the entire request failed a while ago, and this test is just to make sure we exercise that functionality.

As for the mock, that may be valuable when rewriting tests for the directory service module itself, but I think for the user module, it is probably more work since we already catch the error. This is also depending on what Henry does for implementing generic error handling later as well, so this may change in the future, but I just wrote the test for the current version to keep it simple.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11)*

jpevarnek:
:thumbsup: with one quesiton

---

Review status: 2 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[ufo/templates/user_details.html, line 9 \[r7\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8UoCuLBS6Qbj87uWs8:-K8VUPPAMlydHKdH0pM6:-1370417982) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9d65fa2f0844eaa7f83453c7ecbd2267f21f4897/ufo/templates/user_details.html#L9)):*
This seems like I horribly screwed up the implementation, I'm going to try debugging what I did wrong there.

---

*[ufo/user_test.py, line 50 \[r7\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8UpfXlGwtW5jr_VBww:-K8VUjY8_3neVWV4JJjM:-1240381312) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9d65fa2f0844eaa7f83453c7ecbd2267f21f4897/ufo/user_test.py#L50)):*
The private and public key seem like they'd be good to keep accessible in case we need to override verification somewhere, my comment was just that it seems intuitive that our standard key will not be revoked.  Your call on how to handle this.

---

*[ufo/user_test.py, line 110 \[r7\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8UqGN7WV7UiSv_AP7q:-K8VUw4EHnCl2PWudINl:1304636392) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9d65fa2f0844eaa7f83453c7ecbd2267f21f4897/ufo/user_test.py#L110)):*
Okay, makes sense.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-:-K8VV0ZHLTQOZM3leeaW:-1240892515)*

xwsxethan:
Review status: 2 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[ufo/templates/user_details.html, line 9 \[r7\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8UoCuLBS6Qbj87uWs8:-K8VVzD1V0Su9lrQatdy:834371339) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9d65fa2f0844eaa7f83453c7ecbd2267f21f4897/ufo/templates/user_details.html#L9)):*
Okay, I'll leave this as is and let you adjust this implementation once the error's sorted out.

---

*[ufo/user_test.py, line 255 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85ruHpFKHi48qfHA4n:-K8VVsSoqBU-0WXXPqDY:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/986891a8e9f949415152e31e8afa7e3924fa408c/ufo/user_test.py#L255)):*
Done.

---

*[ufo/user_test.py, line 97 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K85kOEaOXzJMmS008ds-r2-97:-K8VVlo0rfpVHHFQPOmN:1689193178) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/bf057009a88c5314a7d9c4db71e276032a20d903/ufo/user_test.py#L97)):*
Sounds good.

---

*[ufo/user_test.py, line 50 \[r7\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11#-K8UpfXlGwtW5jr_VBww:-K8VVa-awleAX4qrcUfC:-1057813307) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9d65fa2f0844eaa7f83453c7ecbd2267f21f4897/ufo/user_test.py#L50)):*
Ehh, I'll just leave it like this for now, and we can always go back and define them as vars again if necessary or just reference them from FAKE_MODEL_USER.property_goes_here.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11)*

xwsxethan:
Reviewed 1 of 3 files at r2, 1 of 2 files at r4, 3 of 3 files at r8.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/11)*

PR #12, opened 2016-01-19
Fold the database setup into app runner.

blueandgold:
This should setup the database if it doesn't already exists when we run the app.  If I understand flask correctly, this should be a beneficial change, and save an extra file/step.  Unless there's a reason why the database setup needs to be in a separate file?

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/12)


xwsxethan:
Thanks for going back to fix Travis (was about to ask for that). This seems fine to me, but I don't know if there was a larger reason for this separation. Perhaps it is necessary for Heroku or something else? :lgtm: but double check with @jpevarnek to be sure.

---

Reviewed 2 of 2 files at r1, 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/12#-:-K8RIUitYVhMmZ2TcA8n:150993478)*

blueandgold:
@jpevarnek  please take a look as well

jpevarnek:
Reason for the separation is so that we are not calling `create_all` if we simply restart the server.  My instinct is that this could have problems, but I don't know for sure.

By the way, for Heroku, we still need the setup_database script.  It does not use `run.py` (that's just a simple wrapper for testing purposes and possibly if deploying without a true app runner)

`Procfile` contains the startup code for Heroku deployments which still tries to use the pyton script to set up the database, it would be great if we could keep the application compatible with Heroku.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/12#-:-K8RNTeRGlm7WwYlCKEB:1896670820)*

goldspin:
Okay, that's good to know.  I'll revert this.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/12#-:-K8RTqcp5CHGknTXadIU:-1155600619)*

PR #14, opened 2016-01-20
Rename some of the functions in the basemodel

jpevarnek:
This renames two of the functions in the Model class and updates them to
take as parameters a commit argument that determines whether the
session will actually be committed immediately following that function.

We now directoly call methods on the query attribute for the models,
allowing us to use get_or_404 and remove a few TODOs from the code

A few fixes were mode to the add_user functionality so that should
actually work (was not intending to do this, noticed some things while
working on it).

@xwsxethan this is going to have conflicts with your pending PR I think, I'll deal with those once it's merged in if that's fine.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/14)

### DIFF
--- a/ufo/models.py
+++ b/ufo/models.py
@@ -5,25 +5,26 @@ from Crypto.PublicKey import RSA
 LONG_STRING_LENGTH = 1024
 
 class Model(db.Model):
+  """Helpful functions for the database models
+
+  Most method implementations are taken from
+  https://github.com/sloria/cookiecutter-flask"""
   __abstract__ = True
 
-  @classmethod
-  def GetAll(cls):
-    return cls.query.all()
+  def update(self, commit=True, **kwargs):
+    for attr, value in kwargs.items():
+      setattr(self, attr, value)
+    return commit and self.save() or self
 
-  @classmethod
-  def GetById(cls, id):
-    return cls.query.get(id)
-
-  def Add(self):
+  def save(self, commit=True):
     db.session.add(self)
-    db.session.commit()
-
+    if commit:
+      db.session.commit()
     return self
 
-  def Delete(self):
+  def delete(self, commit=True):
     db.session.delete(self)
-    db.session.commit()
+    return commit and db.session.commit()
 
 
 class Config(Model):
@@ -54,7 +55,9 @@ class User(Model):
   public_key = db.Column(db.LargeBinary())
   is_key_revoked = db.Column(db.Boolean(), default=False)
 
-  def __init__(self):
+  def __init__(self, **kwargs):
+    super(User, self).__init__(**kwargs)
+
     self.regenerate_key_pair()
 
   @staticmethod
@@ -90,9 +93,3 @@ class ProxyServer(Model):
   name = db.Column(db.String(LONG_STRING_LENGTH))
   ssh_private_key = db.Column(db.String(LONG_STRING_LENGTH))
   fingerprint = db.Column(db.String(LONG_STRING_LENGTH))
-
-  def __init__(self, ip_address, name, ssh_private_key, fingerprint):
-    self.ip_address=ip_address
-    self.name = name
-    self.ssh_private_key = ssh_private_key
-    self.fingerprint = fingerprint
--- a/ufo/user_test.py
+++ b/ufo/user_test.py
@@ -50,8 +50,6 @@ FAKE_MODEL_USER = MagicMock(email=FAKE_EMAILS_AND_NAMES[0]['email'],
                             private_key='private key foo',
                             public_key='public key bar',
                             is_key_revoked=False)
-FAKE_ID = 1000
-
 class UserTest(base_test.BaseTest):
   """Test user class functionality."""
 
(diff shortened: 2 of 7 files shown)


xwsxethan:
Looks fine overall, I just have a few comments.

First, is there any purpose behind starting with caps vs not? For instance, the methods were previously Add and Delete and are now save and delete. I don't really care either way personally, but if there is a greater convention here then we should stick with it. I assumed previously that the caps were to denote something.

Second, the use of assertEquals(foo, True) was for more verbose output in the event of a failure. The stack trace for assertEquals shows False != True and above that will actually show the line it evaluated to False, whereas assertTrue just shows the evaluated version, such as just 'expected True and got False'. I'm fine with using assertTrue since it is obviously cleaner, but wanted to point this out and explain the reasoning. If we feel the more verbose logging is necessary then we should switch those back. Henry may feel more strongly about this.

:lgtm: overall.

---

Reviewed 7 of 7 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/14#-:-K8UtiTv9wM4JmSNZmR2:-882302428)*

jpevarnek:
As far as the naming, I just had never seen methods named starting with a capital letter in Python before.  If there was a reason for that, then we should keep it.  Let's wait and check with @blueandgold 

Huh, hadn't thought of that for `assertEquals`, seems like a bit of a strange thing.  Do you think it would be better if we added a message to the `assertTrue` calls explaining what it was testing specifically?

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/14#-:-K8UxOE9dPoS7ddS6J0I:-1480447266)*

xwsxethan:
If assertTrue supports that then yeah maybe, although that may be more work to write comments for every assertTrue to make it obvious on what conditions the assert acts. As long as there is a line number in the trace, it isn't really that hard to track down either way, but it doesn't help with deciphering what made a value false-y when it should have been truth-y. Just food for though, but I don't have much of a preference so long as we are consistent.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/14#-:-K8V0dV6-xFI0P6amFxX:-588466880)*

goldspin:
:lgtm:  Looks good to me overall.  The name casing is fine with me since that's the external convention, and as long as we stick with it everywhere.

As for the assert statement, yah, from a purist point of view, we should use assertTrue on booleans, and assertEquals when doing actual comparison.  Really, there's nothing that tells which assert statement borks?

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/14#-:-K8VTTst00sBrkBeCMce:571498415)*

jpevarnek:
PTAL (there were a lot of changes with the merge)

---

Review status: 5 of 7 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/14#-:-K8XFklHeUAxrcIcq1tm:-242655735)*

xwsxethan:
Looks good to me. :+1: 

---

Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[ufo/user.py, line 131 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/14#-K8ZohARA6teuzVNdX5X:-K8ZohARA6teuzVNdX5Y:-490100129) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/a3e1f66b93ee3302f60c73c85d3357032ed1bc28/ufo/user.py#L131)):*
Nice!

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/14#-:-K8ZpLEABc6d-I00FxfG:1426114601)*

ufo/models.py
```diff
@@ -5,25 +5,26 @@
...
-
+    if commit:
+      db.session.commit()
     return self
 
-  def Delete(self):
+  def delete(self, commit=True):
     db.session.delete(self)
-    db.session.commit()
+    return commit and db.session.commit()
```
blueandgold: Why is it that in save() there is an if condition on committing, but not so here?
For example:
+    if commit:
 -		 +      db.session.commit()

ufo/user_test.py
```diff
@@ -246,40 +240,42 @@ def testAddUsersPostManualHandler(self, mock_add):
     response = self.client.post(flask.url_for('add_user'), data=data,
                                 follow_redirects=False)
 
+    matched_count = models.User.query.filter_by(email=FAKE_EMAILS_AND_NAMES[0]['email']).count()
```
blueandgold: It would be nice to assert a bit more on the manual add's execution path, rather just a count of something was added.  For example, the data could have user_name and user_email and then test if those are populated into the database correctly.

 user_name = flask.request.form.get('user_name')
 user_email = flask.request.form.get('user_email')

blueandgold:
:+1: Looking at this a bit late.

ufo/user_test.py
jpevarnek: Ah, sorry, bit of clarification on the above line:

It tests that precisely one record now exists in the database that has that email address.

I'd personally argue that this is an improvement over the previous way of testing that it got added which was to make sure that `models.User.Add` got called at least once with no arguments.

ufo/models.py
jpevarnek: It's done in the more verbose way in `save` because we want to return the current object no matter what, this is just a shorter way to write it (and also what I copied from the skeleton as noted above).

I'd be fine with switching it to be more verbose if you would like to do that.

blueandgold: Ah, okay.  Personally, it would be nice to be more verbose to be consistent, and also save some headscratching as we are still learning the framework. :)

ufo/user_test.py
blueandgold: Yes, this definitely is better than before.  I am trying to make this even better, for the similar test for proxy server.  Let me take care of this for you when I get that working.

PR #17, opened 2016-01-21
New Google Directory Service Tests

xwsxethan:
I attempted to utilize the mocking style provided here: https://developers.google.com/api-client-library/python/guide/mocks

However, I could not get GDS to build with the mock http sequence. It may work if we capture an entire request and response into a json file as they suggest, but since we're dealing with the directory service, that would imply that we keep a json file somewhere containing an entire blob for a domain which will then contain PII about all users in the domain. We can of course modify this to be fake data, but overall that approach seems like more work than just mocking out the calls we need directly. Please take a look at how I've done it here and see what you both think, @blueandgold and @jpevarnek .

I'll expand this to tests for the other methods if we agree that this seems feasible. Thanks!

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/17)

### DIFF
--- a/ufo/google_directory_service.py
+++ b/ufo/google_directory_service.py
@@ -5,22 +5,20 @@ import httplib2
 from googleapiclient import discovery
 
 MY_CUSTOMER_ALIAS = 'my_customer'
-
 NUM_RETRIES = 3
-
-VALID_WATCH_EVENTS = ['add', 'delete', 'makeAdmin', 'undelete', 'update']
-
-# TODO(eholder): Write tests for these functions.
 
 class GoogleDirectoryService(object):
 
   """Interact with Google Directory API."""
 
-  def __init__(self, credentials):
+  def __init__(self, credentials, http=None):
     """Create a service object for admin directory services using oauth."""
+    if http is None:
+      http = credentials.authorize(httplib2.Http())
     self.service = discovery.build(serviceName='admin',
                                    version='directory_v1',
-                                   http=credentials.authorize(httplib2.Http()))
+                                   http=http,
+                                   developerKey='myapikey1234')
 
   def GetUsers(self):
     """Get the users of a customer account.
--- a/ufo/google_directory_service.py
+++ b/ufo/google_directory_service.py
@@ -11,14 +11,11 @@ class GoogleDirectoryService(object):
 
   """Interact with Google Directory API."""
 
-  def __init__(self, credentials, http=None):
+  def __init__(self, credentials):
     """Create a service object for admin directory services using oauth."""
-    if http is None:
-      http = credentials.authorize(httplib2.Http())
     self.service = discovery.build(serviceName='admin',
                                    version='directory_v1',
-                                   http=http,
-                                   developerKey='myapikey1234')
+                                   http=credentials.authorize(httplib2.Http()))
 
   def GetUsers(self):
     """Get the users of a customer account.
--- a/ufo/google_directory_service_test.py
+++ b/ufo/google_directory_service_test.py
@@ -3,7 +3,7 @@ from mock import MagicMock
 from mock import patch
 
 import flask
-from googleapiclient import http
+from googleapiclient import discovery
 import json
 import unittest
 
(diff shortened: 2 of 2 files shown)


blueandgold:
:+1:  Looks good to me.  I am okay to mock these out.  :)

jpevarnek:
Reviewed 2 of 2 files at r1.

---

*[ufo/google_directory_service_test.py, line 63 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/17#-K8aqGAipfGtwTA3WRaF:-K8aqGAipfGtwTA3WRaG:307453561) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/cf10d94967943ab59438a6123695ce9b14ede307/ufo/google_directory_service_test.py#L63)):*
So, my interpretation of what this is actually testing is that `GetUsers` precisely returns `distovery.build().users().list().execute()`, does that seem accurate?  Honestly, not sure what we are expecting from this test.

---

*[ufo/google_directory_service_test.py, line 92 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/17#-K8atQRZwQE-mO3S0lRO:-K8atQRZwQE-mO3S0lRP:-1096381740) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/cf10d94967943ab59438a6123695ce9b14ede307/ufo/google_directory_service_test.py#L92)):*
Better name please

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/17)*

xwsxethan:

---

*[ufo/google_directory_service_test.py, line 63 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/17#-K8aqGAipfGtwTA3WRaF:-K8au7yEIdRr0aFyInBN:-24974681) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/cf10d94967943ab59438a6123695ce9b14ede307/ufo/google_directory_service_test.py#L63)):*
Pretty much, yeah. Without being able to hook into a real data source, there isn't much else to assert here. It does also implicitly test that the call does not loop infinitely (since there is a while true). I can expand it to test the scenario with a page token, which may be more useful, but in reality, these methods are all glorified wrappers to call the specific methods we need on the directory API, so there aren't a lot of options without real data. If we're fine with no tests for this, then that works for me too haha.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/17)*

jpevarnek:

---

*[ufo/google_directory_service_test.py, line 63 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/17#-K8aqGAipfGtwTA3WRaF:-K8b4s2yNf1LwseoN4UC:-1009109653) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/cf10d94967943ab59438a6123695ce9b14ede307/ufo/google_directory_service_test.py#L63)):*
I'd be fine either way to be honest.

I think the thing we'd actually want to test for the methods like this is that they correctly do handle the looping since that's about the only place I could imagine these failing horribly.  I'll leave that up to you, no strong objections I guess, I'm just not really looking forward to having to update the tests as we move along...

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/17)*

xwsxethan:
@jpevarnek  PTAL as I updated the tests to only look at paged results and ensuring that they are handled properly. Also randomly updated some things for documentation and cleanliness.

---

Reviewed 1 of 1 files at r2.

---

*[ufo/google_directory_service_test.py, line 92 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/17#-K8atQRZwQE-mO3S0lRO:-K9TUbxkME3lN0lZBT4o:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/cf10d94967943ab59438a6123695ce9b14ede307/ufo/google_directory_service_test.py#L92)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/17#-:-K9TUelnIWNPS_k8r86p:-393529548)*

PR #19, opened 2016-01-22
Migrating Web Driver Functional Tests Over to Flask Version

xwsxethan:
I was able to migrate these fairly easily. The only piece I left out was the login script since we don't have any login and won't need it yet. I updated the tests to use flask.url_for(some-handler-here) instead of the old paths variables so that should be pretty robust.

These can be run manually as before (see the readme in the old repo for instructions) or with the new script I created. The script can be used both to install web driver (via ./web_driver.sh install) and to run the tests (via ./web_driver.sh test server-url-goes-here email-goes-here password-goes-here). Here is an example:

./web_driver.sh install

./web_driver.sh test http://0.0.0.0:5000 ethan@henrychang.mygbiz.com password-goes-here

Note no trailing / in the url as that will cause a problem. Also, since there is no login yet, the email and password fields don't really matter, but I made the script require them for later, so feel free to enter anything in them for now.

EDIT:
Also note, for the tests to execute properly, you will need a server already running and configured. This does not start up a local server for you, though that is something we could add in the future. This means not only doing python run.py (with & appended if you want it to run headless in the background), but going through the flow to manually setup oauth and domain before running the tests. I personally keep an instance configured locally to test with and then just turn it on as needed.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/19)

### DIFF
--- a/tests/ui/base_driver.py
+++ b/tests/ui/base_driver.py
@@ -1,0 +1,11 @@
+"""Base driver module to inherit from."""
+
+class BaseDriver(object):
+
+  """Base driver that will be called from all pages and elements."""
+
+  # pylint: disable=too-few-public-methods
+
+  def __init__(self, driver):
+    """Create the base driver object."""
+    self.driver = driver
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,4 @@ venv/*
 *.pyc
 ufo/static/bower_components/*
 node_modules/*
-lib/*
+chrome_driver/*
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ venv/*
 *.pyc
 ufo/static/bower_components/*
 node_modules/*
+lib/*
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,6 +13,7 @@ pyasn1==0.1.9
 pyasn1-modules==0.0.8
 pycrypto==2.6.1
 rsa==3.2.3
+selenium==2.49.2
 simplejson==3.8.1
 six==1.10.0
 SQLAlchemy==1.0.11
--- a/tests/ui/base_test.py
+++ b/tests/ui/base_test.py
@@ -6,8 +6,7 @@ class BaseTest(unittest.TestCase):
 
   """Base test class to inherit from."""
 
-  def __init__(self, methodName='runTest', args=None, kwargs=None):
+  def __init__(self, methodName='runTest', args=None, **kwargs):
     """Create the base test object for others to inherit."""
-    super(BaseTest, self).__init__(methodName)
+    super(BaseTest, self).__init__(methodName, **kwargs)
     self.args = args
-    self.kwargs = kwargs
--- a/tests/ui/base_test.py
+++ b/tests/ui/base_test.py
@@ -1,0 +1,12 @@
+"""Test base module functionality."""
+
+import unittest
+
+class BaseTest(unittest.TestCase):
+
+  """Base test class to inherit from."""
+
+  def __init__(self, methodName='runTest', args=None):
+    """Create the base test object for others to inherit."""
+    super(BaseTest, self).__init__(methodName)
+    self.args = args
--- a/tests/ui/base_test.py
+++ b/tests/ui/base_test.py
@@ -6,7 +6,8 @@ class BaseTest(unittest.TestCase):
 
   """Base test class to inherit from."""
 
-  def __init__(self, methodName='runTest', args=None):
+  def __init__(self, methodName='runTest', args=None, kwargs=None):
     """Create the base test object for others to inherit."""
     super(BaseTest, self).__init__(methodName)
     self.args = args
+    self.kwargs = kwargs
--- a/tests/ui/landing_page.py
+++ b/tests/ui/landing_page.py
@@ -1,0 +1,24 @@
+"""Landing page module for testing."""
+from base_driver import BaseDriver
+from sidebar import Sidebar
+
+from selenium.webdriver.common.by import By
+
+class LandingPage(BaseDriver):
+
+  """Home page action methods and locators."""
+
+  TITLE = (By.TAG_NAME, 'h2')
+  INSTRUCTION = (By.TAG_NAME, 'h4')
+
+  def GetTitle(self):
+    """Get the title element on the landing page."""
+    return self.driver.find_element(*self.TITLE)
+
+  def GetInstruction(self):
+    """Get the instructions element on the landing page."""
+    return self.driver.find_element(*self.INSTRUCTION)
+
+  def GetSidebar(self):
+    """Get the sidebar element on the landing page."""
+    return self.driver.find_element(*Sidebar.SIDEBAR)
--- a/tests/ui/landing_page_test.py
+++ b/tests/ui/landing_page_test.py
@@ -20,6 +20,9 @@ class LandingPageTest(BaseTest):
 
   def testLandingPage(self):
     """Test the landing page."""
+    # TODO(eholder): Improve the checks here to be based on something more
+    # robust, such as the presence of element id's or that the page renders
+    # as expected, since this text can change in the future and is not i18ned.
     title = u'Uproxy for Organizations Management Server'
     instruction = ('Click one of the links on the side to login and '
                    'administer the server.')
--- a/tests/ui/landing_page_test.py
+++ b/tests/ui/landing_page_test.py
@@ -1,0 +1,40 @@
+"""Test langing page module functionality."""
+import unittest
+
+from base_test import BaseTest
+from landing_page import LandingPage
+from test_config import CHROME_DRIVER_LOCATION
+
+from selenium import webdriver
+
+
+class LandingPageTest(BaseTest):
+
+  """Test landing page functionality."""
+
+  def setUp(self):
+    """Setup for test methods."""
+    self.driver = webdriver.Chrome(CHROME_DRIVER_LOCATION)
+    # TODO(eholder) Re-enable this once we have a login module again.
+    # LoginPage(self.driver).Login(self.args)
+
+  def testLandingPage(self):
+    """Test the landing page."""
+    title = u'Uproxy for Organizations Management Server'
+    instruction = ('Click one of the links on the side to login and '
+                   'administer the server.')
+
+    self.driver.get(self.args.server_url)
+
+    landing_page = LandingPage(self.driver)
+    self.assertEquals(title, landing_page.GetTitle().text)
+    self.assertEquals(instruction, landing_page.GetInstruction().text)
+    self.assertIsNotNone(landing_page.GetSidebar())
+
+  def tearDown(self):
+    """Teardown for test methods."""
+    self.driver.quit()
+
+
+if __name__ == '__main__':
+  unittest.main()
--- a/tests/ui/sidebar.py
+++ b/tests/ui/sidebar.py
@@ -1,0 +1,27 @@
+"""Sidebar module to get links for testing."""
+from base_driver import BaseDriver
+
+from selenium.webdriver.common.by import By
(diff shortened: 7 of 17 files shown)


xwsxethan:
@jpevarnek PTAL as well por favor

jpevarnek:
I just skimmed this, but from what I see, it looks like there's a lot of checking to see that the wording is exactly what it is now.  I'm honestly just a bit worried that we'll end up getting feedback from someone that a bit of text is not exactly the best way to say things (there are a lot of points right now where we just have filler text) and end up having to change it in multiple tests as well as the code, which is really the point at which you stop paying attention to tests.  If that's what we're doing, I think we'd be better using an image-comparison service that gives us an easy way to say that the change was expected.

I'm looking specifically at the instruction test when I make the above comment.

In general, with testing, I like to think about what my likely reaction will be if I see that a specific test broke.  If the reaction is "Oh, I probably updated something and forgot to update the test", I treat that as a big mental red flag that the test probably should be improved.

Also, any thoughts about moving our current tests into the `tests` directory?

---

Reviewed 5 of 14 files at r1.
Review status: 5 of 14 files reviewed at latest revision, 3 unresolved discussions.

---

*[.gitignore, line 8 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-K8b7deoBhKGFD751FNI:-K8b7depdq4cDznBOtlx:2135449273) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/76cf7215f72c4744089cdb3a2d03e82d622b43d2/.gitignore#L8)):*
What is this for?

---

*[tests/ui/base_test.py, line 9 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-K8b8-GE9z1kS_V0qHK0:-K8b7yW6QhD6bMcJMe_J:996878377) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/76cf7215f72c4744089cdb3a2d03e82d622b43d2/tests/ui/base_test.py#L9)):*
Might be useful to also pass kwargs to the constructor in case we wanted to use the other functionality.

---

*[ufo/setup.py, line 16 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-K8b8cX2s8zUVU6ScIZN:-K8b8cX2s8zUVU6ScIZO:-228146910) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/76cf7215f72c4744089cdb3a2d03e82d622b43d2/ufo/setup.py#L16)):*
See redirection behaviour at http://flask.pocoo.org/docs/0.10/quickstart/, I think we want to keep the trailing `/` unless there is a concrete reason why it is causing problems.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-:-K8b94BF6snOWG9WovK5:-1166104203)*

tests/ui/base_driver.py
```diff
@@ -0,0 +1,11 @@
+"""Base driver module to inherit from."""
+
+class BaseDriver(object):
+
```
blueandgold: nit: drop this whitespace

blueandgold:
:lgtm:   Please see my comment about not having another wrapper if at all possible.

Yes, checking on the string can probably be improved upon.  We can check on whether the string element exists. i.e. by using shareable id attribute.  But it's outside the scope of this PR, so I would just put it as a TODO for when we work on the actual UI mock.

---

Review status: 5 of 14 files reviewed at latest revision, 4 unresolved discussions.

---

*[web_driver.sh, line 3 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-K8eyYkpgSrQ4_wmYJts:-K8eyYkqK82Kcxsg-WzY:1605562815) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/76cf7215f72c4744089cdb3a2d03e82d622b43d2/web_driver.sh#L3)):*
It's a good idea to have a separate installer for the webdriver stuff.  However, I wonder if we really need another wrapper on top of a wrapper to run the tests.  Can the env not be properly set in ui_test_suite.py and the new help text to be there as well?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-:-K8ewyLBKGiY498Z-zaA:-1197730485)*

xwsxethan:
That makes sense for the text elements. I agree, the checking here could be more robust based on the elements appearing and things rendering all the way. As per Henry's suggestion, I've left a TODO for that. I already have a work item for improving these tests anyway, so I'll make that part of it. For now, I just wanted to migrate them over so that we have something.

---

Reviewed 9 of 14 files at r1, 8 of 8 files at r2.

---

*[.gitignore, line 8 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-K8b7deoBhKGFD751FNI:-K8u8NgYNUQmBHNNO1LX:175724242) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/76cf7215f72c4744089cdb3a2d03e82d622b43d2/.gitignore#L8)):*
This is where the chromedriver frameworks gets installed. That's what we had used before with the appengine project, so I just kept it the same. I can put it somewhere else if we like.

---

*[tests/ui/base_test.py, line 9 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-K8b8-GE9z1kS_V0qHK0:-K8uBnzXmEQTBc5HCL5u:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/76cf7215f72c4744089cdb3a2d03e82d622b43d2/tests/ui/base_test.py#L9)):*
Done.

---

*[web_driver.sh, line 3 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-K8eyYkpgSrQ4_wmYJts:-K8uAjblot97kk5fWemI:-1314203471) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/76cf7215f72c4744089cdb3a2d03e82d622b43d2/web_driver.sh#L3)):*
It definitely still works when run manually as before. There was no error with that means of executing it, but since I was making the script for installation, it felt like I should have an option to run it as well. Either method is perfectly valid and performs identically, so I'd like to keep this around, but don't feel as though you have to use it.

---

*[ufo/setup.py, line 16 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-K8b8cX2s8zUVU6ScIZN:-K8u9F6Pw2mWDX882yeQ:544484635) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/76cf7215f72c4744089cdb3a2d03e82d622b43d2/ufo/setup.py#L16)):*
Hmmmm, interesting. I was changing it to make it consistent with how we previously had things in appengine, which was no trailing slash. Their reasoning seems fine for changing it, though I don't think it will be entirely necessary in our context.

The problem with the trailing slash was that it didn't match what is defined in the sidebar. The sidebar defined in sidebar.js used urls with no trailing slash (except for the root handler, which was literally just "/"). This causes the sidebar to not show the current page in bold when you navigate to it (it is expecting you to be on /user, but flask has put you on /user/ so it thinks you're on an unspecified subpage). Similarly, in the functional tests, it expects the href to redirect you to the url_for(setup) which was /setup/ while the href actually points to just /setup. It doesn't really matter which format we go with, just so long as they match up. I can change them to include the trailing slash throughout.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-:-K8uBE-BnLCZIAYW20zy:-874407752)*

jpevarnek:
Reviewed 1 of 14 files at r1, 8 of 8 files at r2.

---

*[.gitignore, line 8 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-K8b7deoBhKGFD751FNI:-K8uVsCw2-tA7HZCy0Ww:-1701742389) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/76cf7215f72c4744089cdb3a2d03e82d622b43d2/.gitignore#L8)):*
I think that was just done because we were downloading other libraries there as well.  I don't think `lib` makes total sense, but I'll totally defer to you on the naming.

---

*[tests/ui/base_test.py, line 9 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-K8b8-GE9z1kS_V0qHK0:-K8uW5ASS8ke7giaO5PW:563276793) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/76cf7215f72c4744089cdb3a2d03e82d622b43d2/tests/ui/base_test.py#L9)):*
I'm sorry, for kwargs, I meant taking the special `**kwargs` parameter and passing that through to the base class (similar to http://stackoverflow.com/questions/5031711/python-cleanest-way-to-override-init-where-an-optional-kwarg-must-be-used)

With your change, what are you planning on using `self.kwargs` for?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19)*

xwsxethan:
Review status: 12 of 16 files reviewed at latest revision, 3 unresolved discussions.

---

*[.gitignore, line 8 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-K8b7deoBhKGFD751FNI:-K8umTh4z_OrUDsSOQos:1761059787) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/76cf7215f72c4744089cdb3a2d03e82d622b43d2/.gitignore#L8)):*
I renamed the directory to chrome_driver and updated this and the script if that works.

---

*[tests/ui/base_test.py, line 9 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-K8b8-GE9z1kS_V0qHK0:-K8umRIUQS3GY5mmGP8Z:-698231926) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/76cf7215f72c4744089cdb3a2d03e82d622b43d2/tests/ui/base_test.py#L9)):*
Better?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19)*

jpevarnek:
:thumbsup: at your discretion

---

Reviewed 4 of 4 files at r3.

---

*[tests/ui/test_config.py, line 7 \[r3\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-K8yQWoIWjzDAXZwhnhJ:-K8yQWoIWjzDAXZwhnhK:1806652467) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b2c0e15dda199fa3dccce64871f7dc608c359a5d/tests/ui/test_config.py#L7)):*
Are both directories necessary?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-:-K8yQhUSIDVHVO6HFh_a:963321807)*

xwsxethan:

---

*[tests/ui/test_config.py, line 7 \[r3\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19#-K8yQWoIWjzDAXZwhnhJ:-K8yUMfd4jdszDqxKPGO:-2096967276) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b2c0e15dda199fa3dccce64871f7dc608c359a5d/tests/ui/test_config.py#L7)):*
The second one (chromedriver, not chrome_driver) is not a directory; it is the actual executable. I'm not sure why it doesn't have a file ending though. That's just how it comes when it is unzipped. I could just put it in the top level directory straight up if that is cleaner.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/19)*

PR #20, opened 2016-01-22
Improve and Cleanup the User Test

blueandgold:
- better assert that users are added correctly
- factor out fake user creation
- better python looping
- small renaming

I got tired at this point.  Please let me know if you see something else to improve.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/20)

### DIFF
--- a/ufo/user_test.py
+++ b/ufo/user_test.py
@@ -232,35 +232,33 @@ class UserTest(base_test.BaseTest):
     self.assertEquals(len(FAKE_EMAILS_AND_NAMES), len(users_in_db))
 
     for fake_email_and_name in FAKE_EMAILS_AND_NAMES:
-      user_in_db = models.User.query.filter_by(
-          email=fake_email_and_name['email']).one_or_none()
+      query = models.User.query.filter_by(email=fake_email_and_name['email'])
+      user_in_db = query.one_or_none()
       self.assertEqual(fake_email_and_name['name'], user_in_db.name)
 
     self.assert_redirects(response, flask.url_for('user_list'))
 
   def testAddUsersPostManualHandler(self):
     """Test add users manually calls to insert the specified user."""
-    user = self._CreateFakeUser()
     data = {}
     data['manual'] = True
-    data['user_email'] = user.email
-    data['user_name'] = user.name
+    data['user_email'] = FAKE_EMAILS_AND_NAMES[0]['email']
+    data['user_name'] = FAKE_EMAILS_AND_NAMES[0]['name']
 
     response = self.client.post(flask.url_for('add_user'), data=data,
                                 follow_redirects=False)
 
-    user_in_db = models.User.query.filter_by(email=user.email).one_or_none()
+    query = models.User.query.filter_by(email=FAKE_EMAILS_AND_NAMES[0]['email'])
+    user_in_db = query.one_or_none()
     self.assertIsNotNone(user_in_db)
-    self.assertEqual(user.email, user_in_db.email)
-    self.assertEqual(user.name, user_in_db.name)
+    self.assertEqual(FAKE_EMAILS_AND_NAMES[0]['name'], user_in_db.name)
 
     self.assert_redirects(response, flask.url_for('user_list'))
 
   @patch('flask.render_template')
   def testUserDetailsGet(self, mock_render_template):
     """Test the user details handler calls to render a user's information."""
-    user = self._CreateFakeUser()
-    user.save()
+    user = self._CreateAndSaveFakeUser()
     mock_render_template.return_value = ''
 
     response = self.client.get(flask.url_for('user_details', user_id=user.id))
@@ -276,8 +274,7 @@ class UserTest(base_test.BaseTest):
     fake_ip = '0.1.2.3'
     proxy_server = models.ProxyServer(ip_address=fake_ip)
     proxy_server.save()
-    user = self._CreateFakeUser()
-    user.save()
+    user = self._CreateAndSaveFakeUser()
     mock_render_template.return_value = ''
 
     response = self.client.get(flask.url_for('user_details', user_id=user.id))
@@ -299,8 +296,7 @@ class UserTest(base_test.BaseTest):
 
   def testDeleteUserPostHandler(self):
     """Test the delete user handler calls to delete the specified user."""
-    user = self._CreateFakeUser()
-    user.save()
+    user = self._CreateAndSaveFakeUser()
     user_id = user.id
 
     response = self.client.post(flask.url_for('delete_user', user_id=user_id),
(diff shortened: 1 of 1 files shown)


blueandgold:
@jpevarnek please take a look at well

xwsxethan:
Looks good overall. The reason for looping was because I previously had two arrays before I created one dictionary that combined them, and I had never updated the loops since they still worked. This way with a foreach is definitely cleaner.

I had a question about what we assert on for rendered html vs just what render is called with, so we may want to discuss that.

---

Reviewed 1 of 1 files at r1.

---

*[ufo/user_test.py, line 85 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/20#-K8apOfg4FVf5SSvD97g:-K8apOfg4FVf5SSvD97h:-827822842) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/81b48f414452312dd7ac7be813b90c3c254c118b/ufo/user_test.py#L85)):*
The test here was intentionally avoiding checking what shows up in the html, but rather just checking what we expect from the pure python side.

I certainly don't mind doing it this way or the other, but we should just be consistent. The other tests just assert that render was called with foo and bar parameters, so we'll want to change those as well if we want to assert based on the returned html. What do you think?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/20#-:-K8aqEQnqy4Pdbbpubq3:-1205127346)*

blueandgold:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[ufo/user_test.py, line 85 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/20#-K8apOfg4FVf5SSvD97g:-K8asy5sS8Xtfuz9mHed:1139084508) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/81b48f414452312dd7ac7be813b90c3c254c118b/ufo/user_test.py#L85)):*
Rolling this back to the mock as this method is really small, without any other operations.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/20)*

jpevarnek:
Reviewed 1 of 1 files at r2.

---

*[ufo/user_test.py, line 243 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/20#-K8bB2xsDcLjAQGL1KTq:-K8bB2xsDcLjAQGL1KTr:-581994600) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/cae30e91d496fee8d80947929f227ef694099068/ufo/user_test.py#L243)):*
Same comment as the previous test where it's terrifying to me that we're creating a model for no reason other than to get the data from the default.  If we want to check off of default data and not use a model, I'd much prefer we refer to the data directly.

---

*[ufo/user_test.py, line 254 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/20#-K8bBJKeNUct3_Pw20Vf:-K8bBJKeNUct3_Pw20Vg:-1042091780) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/cae30e91d496fee8d80947929f227ef694099068/ufo/user_test.py#L254)):*
This is trivially true from the query.

---

*[ufo/user_test.py, line 344 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/20#-K8bBQojDr4G9D2ylJJj:-K8bAxbZKwwPL93nEwiv:-174150662) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/cae30e91d496fee8d80947929f227ef694099068/ufo/user_test.py#L344)):*
I think I'd feel a lot better if this was `return models.User(...).save()`

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/20)*

blueandgold:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

*[ufo/user_test.py, line 243 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/20#-K8bB2xsDcLjAQGL1KTq:-K8e_nkLzf1TgkgWAQl6:-770223885) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/cae30e91d496fee8d80947929f227ef694099068/ufo/user_test.py#L243)):*
Done.  It was to make the code here simpler by reusing the method to create a user from the existing data.

---

*[ufo/user_test.py, line 254 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/20#-K8bBJKeNUct3_Pw20Vf:-K8eaJ8JladQuuMTb5BM:-1604963961) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/cae30e91d496fee8d80947929f227ef694099068/ufo/user_test.py#L254)):*
Done.  Dropping this.  :)

---

*[ufo/user_test.py, line 344 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/20#-K8bBQojDr4G9D2ylJJj:-K8ebS5idrPtaeAUC7QY:-2115856614) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/cae30e91d496fee8d80947929f227ef694099068/ufo/user_test.py#L344)):*
Done. I moved the save() step here, but will need to return a user instead.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/20)*

jpevarnek:
One tiny question from me (just want to make sure you saw it, no change required)

:thumbsup: at your discretion

---

Reviewed 1 of 1 files at r3.

---

*[ufo/user_test.py, line 344 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/20#-K8bBQojDr4G9D2ylJJj:-K8efkj1fDcSKE4Uojbr:1868414176) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/cae30e91d496fee8d80947929f227ef694099068/ufo/user_test.py#L344)):*
Note that the `user.save()` returns `user.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/20#-:-K8eg4ckDOMZHINcek5d:1859184791)*

blueandgold:

---

*[ufo/user_test.py, line 344 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/20#-K8bBQojDr4G9D2ylJJj:-K8erKMZLPAU8AmIMKY2:1700096137) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/cae30e91d496fee8d80947929f227ef694099068/ufo/user_test.py#L344)):*
Nice catch!!!  It's working now, not sure why it wasn't when I tried it earlier.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/20)*

PR #22, opened 2016-01-28
Refactoring web driver tests overall and expanding user tests

xwsxethan:
[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/22)

### DIFF
--- a/tests/ui/base_driver.py
+++ b/tests/ui/base_driver.py
@@ -1,11 +1,52 @@
 """Base driver module to inherit from."""
 
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as EC
+from selenium.webdriver.support.ui import WebDriverWait
+
 class BaseDriver(object):
 
   """Base driver that will be called from all pages and elements."""
 
-  # pylint: disable=too-few-public-methods
+  SIDEBAR = (By.TAG_NAME, 'ufo-sidebar')
+  GENERIC_DIV = (By.TAG_NAME, 'div')
 
   def __init__(self, driver):
-    """Create the base driver object."""
+    """Create the base driver object.
+
+    Args:
+      driver: A webdriver instance that can be used to query the dom.
+    """
     self.driver = driver
+
+
+  def GetLink(self, link_locator):
+    """Get an element in the page based on the link locator given.
+
+    Args:
+      link_locator: A tuple of By.SomeProperty and a matching property.
+
+    Returns:
+      A link element if found.
+    """
+    return WebDriverWait(self.driver, 10).until(
+        EC.visibility_of_element_located(((link_locator))))
+
+  def GetElement(self, element_locator):
+    """Get an element in the page based on the element locator given.
+
+    Args:
+      element_locator: A tuple of By.SomeProperty and a matching property.
+
+    Returns:
+      A generic element if found.
+    """
+    return self.driver.find_element(*element_locator)
+
+  def GetSidebar(self):
+    """Get the sidebar element on the user page.
+
+    Returns:
+      The sidebar element for any page.
+    """
+    return self.driver.find_element(*BaseDriver.SIDEBAR)
--- a/tests/ui/base_driver.py
+++ b/tests/ui/base_driver.py
@@ -8,7 +8,6 @@ class BaseDriver(object):
 
   """Base driver that will be called from all pages and elements."""
 
-  SIDEBAR = (By.TAG_NAME, 'ufo-sidebar')
   GENERIC_DIV = (By.TAG_NAME, 'div')
 
   def __init__(self, driver):
@@ -42,11 +41,3 @@ class BaseDriver(object):
       A generic element if found.
     """
     return self.driver.find_element(*element_locator)
-
-  def GetSidebar(self):
-    """Get the sidebar element on the user page.
-
-    Returns:
-      The sidebar element for any page.
-    """
-    return self.driver.find_element(*BaseDriver.SIDEBAR)
--- a/tests/ui/base_test.py
+++ b/tests/ui/base_test.py
@@ -2,6 +2,11 @@
 
 import unittest
 
+from test_config import CHROME_DRIVER_LOCATION
+from ufo import app
+
+from selenium import webdriver
+
 class BaseTest(unittest.TestCase):
 
   """Base test class to inherit from."""
@@ -10,3 +15,18 @@ class BaseTest(unittest.TestCase):
     """Create the base test object for others to inherit."""
     super(BaseTest, self).__init__(methodName, **kwargs)
     self.args = args
+
+  def setUp(self):
+    """Setup for test methods."""
+    self.driver = webdriver.Chrome(CHROME_DRIVER_LOCATION)
+    # TODO(eholder) Re-enable this once we have a login module again.
+    # LoginPage(self.driver).Login(self.args)
+
+  def setContext(self):
+    """Set context as test_request_context so we can use flask.url_for."""
+    self.context = app.test_request_context()
+    self.context.push()
+
+  def tearDown(self):
+    """Teardown for test methods."""
+    self.driver.quit()
--- a/tests/ui/landing_page.py
+++ b/tests/ui/landing_page.py
@@ -1,9 +1,9 @@
 """Landing page module for testing."""
-from sidebar import PageWithSidebar
+from layout import UfOPageLayout
 
 from selenium.webdriver.common.by import By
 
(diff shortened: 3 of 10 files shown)


xwsxethan:
@jpevarnek PTAL as well. I may add more to this PR, but wanted to get a first look at the refactoring. Thanks!

tests/ui/base_driver.py
```diff
@@ -1,11 +1,52 @@
...
+
+    Args:
+      element_locator: A tuple of By.SomeProperty and a matching property.
+
+    Returns:
+      A generic element if found.
+    """
+    return self.driver.find_element(*element_locator)
+
+  def GetSidebar(self):
```
blueandgold: Can you explain why the GetSidebar() should be here?  It obviously is a great idea for generic GetElement() and GetLink(), but not quite for something specific like this.

In other words, can we test with something like:
landing_page.GetElement(sidebar.SIDEBAR)?
where SIDEBAR = (By.TAG_NAME, 'ufo-sidebar') is in sidebar.py?

blueandgold:
The refactoring here is great overall!  One comment, as well as we should wait to expand the tests until we get the new UI mocks.

---

Review status: 0 of 9 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/22#-:-K99BnJifKzj6r7q5FOB:351396629)*

xwsxethan:
Review status: 0 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/ui/base_driver.py, line 46 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/22#-K9960EjlhV51SaSW59D-r1-46:-K99CkgKkJGdBZhKGxuF:-1995038202) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9ecb76881fb1e7c9a91ce7c20418f1307755ff58/tests/ui/base_driver.py#L46)):*
That's a great question. I was a bit undecided on this myself. My reasoning was basically that the sidebar is on every page and has to be tested (at least sanity tested for presence) on every page, so it felt like something that could be easily put in this parent (BaseDriver). It also felt weird to me that every other page had to import sidebar in addition to inherit from BaseDriver so they could do the one sanity check that the sidebar is present. Piggybacking off of the existing inheritance felt like a little more straightforward approach. Thus, I moved it here.

I agree though that this doesn't feel entirely correct either. Truthfully, the best way would be to have sidebar (probably renamed as base_page_with_sidebar) inherit from base to provide its functionality and then have every other page inherit from that so the structure more closely mirrors the underlying representation.

Any of these approaches will work however, it is just a matter of preference. If we feel really strongly about any one approach, then I'll do that since it is only a few lines regardless of the approach. Thanks for the question as it is not necessarily an obvious benefit.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/22)*

blueandgold:
Review status: 0 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/ui/base_driver.py, line 46 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/22#-K9960EjlhV51SaSW59D-r1-46:-K99GObvz2R0_fBXFoVV:-1660791909) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9ecb76881fb1e7c9a91ce7c20418f1307755ff58/tests/ui/base_driver.py#L46)):*
I think we should abide by the page object model as below.  In other words, sidebar.py should be the right home for SIDEBAR = (By.TAG_NAME, 'ufo-sidebar').  If someday this ever needs to be changed, the person would intuitively go to sidebar.py first.

https://code.google.com/p/selenium/wiki/PageObjects
"Facing away from the developer, they should be the only thing that has a deep knowledge of the structure of the HTML of a page (or part of a page)"

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/22)*

xwsxethan:
Reviewed 4 of 9 files at r1, 5 of 5 files at r2.

---

*[tests/ui/base_driver.py, line 46 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/22#-K9960EjlhV51SaSW59D-r1-46:-K9c2gHQ90LHouuQ880c:-1983551907) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9ecb76881fb1e7c9a91ce7c20418f1307755ff58/tests/ui/base_driver.py#L46)):*
What do you think of this way: base_driver is the base page with core functionality; sidebar inherits from base_driver and provides its own functionality (as the class PageWithSidebar); and all other pages inherit from PageWithSidebar so they can easily request the sidebar to verify that it is not None as well as still use the base_driver methods for GetLink and GetElement.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/22)*

blueandgold:

---

*[tests/ui/base_driver.py, line 46 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/22#-K9960EjlhV51SaSW59D-r1-46:-K9c9rPSlMakwrKsBwjo:1448033192) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9ecb76881fb1e7c9a91ce7c20418f1307755ff58/tests/ui/base_driver.py#L46)):*
Hmmm.  It is a bit odd that we are now creating PageWithSidebar class in sidebar.py.  Which leads to these odd statements:
sidebar = PageWithSidebar(self.driver)
class LandingPage(PageWithSidebar)

I am also concerned if there are more objects which also might also need to be inherited this way, e.g. header, footer, even searchbox(?).

So I think, a much simpler & cleaner way is just to remove GetSidebar() from user_page.py  as is the case now.  Then in user_page_test.py, we can just do the following:

from sidebar import Sidebar
self.assertIsNotNone(self.driver.find_element(*Sidebar.SIDEBAR)))

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/22)*

xwsxethan:
Review status: 5 of 10 files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/ui/base_driver.py, line 46 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/22#-K9960EjlhV51SaSW59D-r1-46:-K9cHt2hkJhZmGnjWp3N:313183950) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9ecb76881fb1e7c9a91ce7c20418f1307755ff58/tests/ui/base_driver.py#L46)):*
I'm not sure why the second statement is odd. A LandingPage is/extends a PageWithSidebar.

Anyway the point you bring up is perhaps more of a reason to have an intermediary like this (though the name I provided can be improved). It would seem like a lot of duplication for every separate page to import sidebar, header, footer, searchbox, and any other common elements. There should basically be a layout or template of some sort defined that provides the functionality which every other page (user, proxy server, landing, setup, etc.) then has access to.

It's exactly what we do in the templates themselves with layout.html, because it formalizes the inheritance structure since all pages will share that same model. Then we can just change that intermediary layout once to apply it to all sub pages. This will also be useful if we ever want to assert on where things specifically go on the page, such as the sidebar should be on the left side and be x by y pixels in size.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/22)*

blueandgold:
:+1:  Looks nice now!

---

Review status: 5 of 10 files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/ui/base_driver.py, line 46 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/22#-K9960EjlhV51SaSW59D-r1-46:-K9cORAsc0mIrNBKGY9D:-1262151859) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/9ecb76881fb1e7c9a91ce7c20418f1307755ff58/tests/ui/base_driver.py#L46)):*
Cool, having a UfOPageLayout class to hold these makes everything look much better now.

As for the second statement, it was to illustrate the naming can be better, that we are not really inheriting a page.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/22#-:-K9cVX-uoZBEzETxJSuM:-1471663709)*

PR #23, opened 2016-01-30
Use textarea instead of paper-textarea in the proxy server form.

blueandgold:
This is a temporary fix for the issue that flask is unable to handle the data in paper-textarea.  I say temporary because it's better to wait the UI mocks to land and address this for good at that time.  But for now, this gets the form back into a working state.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/23)


blueandgold:
@jpevarnek PTAL as well, thanks!

jpevarnek:
See #24 for an alternative, your call on which direction is easier

blueandgold:
Your PR is toally cool.  Will drop this.

PR #26, opened 2016-02-03
Create Managed Chrome Policy from the Management Server

xwsxethan:
Currently, this reads values from the database and generates the JSON file which would be uploaded to admin console and pushed down to the uProxy app and extension. It surfaces the JSON file as a new Chrome Policy handler (at /chromepolicy/), which displays the file in a card (for copy pasting) and provides a download link users can click to have the file automatically downloaded for them.

TODO's:

<del>Add support for modifying the new config values (whether to enforce the proxy server check and whether to enforce the network jail prior to Google login) on the management server .</del>

Possibly move the current functionality (and new config value edit functionality) to be a sub-page under the setup handler rather than its own handler and link on the sidebar.

<del>Add text on the handler page to explain what chrome policy is for and where it should go, as well as a link to admin console.</del>

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/26)

### DIFF
--- a/ufo/chrome_policy.py
+++ b/ufo/chrome_policy.py
@@ -1,0 +1,38 @@
+"""The module for generating and outputing chrome policy."""
+
+from . import app, setup_required
+
+import flask
+import json
+
+import models
+
+
+def _MakeChromePolicyJson():
+  """Generate the json string of chrome policy based on values in the database.
+
+  This policy string has the following form:
+
+  Returns:
+    A json string of current chrome policy.
+  """
+  proxy_servers = models.ProxyServer.query.all()
+  proxy_server_public_keys = [s._MakePublicKey() for s in proxy_servers]
+
+  config = models.Config.query.get(0)
+
+  policy_dictionary = {
+      "proxy_server_keys": proxy_server_public_keys,
+      "enforce_proxy_server_validity": config.proxy_server_validity,
+      "enforce_network_jail": config.network_jail_until_google_auth,
+  }
+
+  return json.dumps(policy_dictionary)
+
+
+@app.route('/chromepolicy/')
+@setup_required
+def chrome_policy():
+  policy_json = _MakeChromePolicyJson()
+
+  return flask.render_template('chrome_policy.html', policy_json=policy_json)
--- a/ufo/chrome_policy.py
+++ b/ufo/chrome_policy.py
@@ -36,3 +36,9 @@ def chrome_policy():
   policy_json = _MakeChromePolicyJson()
 
   return flask.render_template('chrome_policy.html', policy_json=policy_json)
+
+
+@app.route('/chromepolicy/download/')
+@setup_required
+def chrome_policy_download():
+  return _MakeChromePolicyJson()
--- a/ufo/chrome_policy.py
+++ b/ufo/chrome_policy.py
@@ -5,7 +5,7 @@ from . import app, setup_required
 import flask
 import json
 
-import models
+import ufo.models
 
 
 def _make_chrome_policy_json():
@@ -16,10 +16,10 @@ def _make_chrome_policy_json():
   Returns:
     A json string of current chrome policy.
   """
-  proxy_servers = models.ProxyServer.query.all()
-  proxy_server_public_keys = [s._MakePublicKey() for s in proxy_servers]
+  proxy_servers = ufo.models.ProxyServer.query.all()
+  proxy_server_public_keys = [s.make_public_key() for s in proxy_servers]
 
-  config = models.Config.query.get(0)
+  config = ufo.models.Config.query.get(0)
 
   policy_dictionary = {
       "proxy_server_keys": proxy_server_public_keys,
@@ -39,7 +39,7 @@ def display_chrome_policy():
     The rendered chrom_policy.html template with policy values as variables.
   """
   policy_json = _make_chrome_policy_json()
-  config = models.Config.query.get(0)
+  config = ufo.models.Config.query.get(0)
 
   return flask.render_template(
       'chrome_policy.html', policy_json=policy_json,
@@ -68,7 +68,7 @@ def edit_policy_config():
     A redirect back to display chrome policy with will display the new values.
   """
 
-  config = models.Config.query.get(0)
+  config = ufo.models.Config.query.get(0)
 
   # TODO(eholder): The Polymer html for sending toggle button values as inputs
   # really sucks. I basically have to parse it out manually in JS then set it
--- a/ufo/chrome_policy.py
+++ b/ufo/chrome_policy.py
@@ -75,19 +75,10 @@ def edit_policy_config():
 
   config = ufo.get_user_config()
 
-  # TODO(eholder): The Polymer html for sending toggle button values as inputs
-  # really sucks. I basically have to parse it out manually in JS then set it
-  # to a hidden input value. If we used iron-form, that may help with setting
-  # those as inputs on its own. That introduces a problem with refreshing the
-  # page on response or writing the response back into the UI for an update...
-  # TODO(eholder): There has to be a better way to convert strings to bools...
   proxy_server_string = flask.request.form.get('enforce_proxy_server_validity')
   network_jail_string = flask.request.form.get('enforce_network_jail')
-
-  config.proxy_server_validity = (proxy_server_string == 'true' or
-                                  proxy_server_string == 'True')
-  config.network_jail_until_google_auth = (network_jail_string == 'true' or
-                                           network_jail_string == 'True')
+  config.proxy_server_validity = json.loads(proxy_server_string)
+  config.network_jail_until_google_auth = json.loads(network_jail_string)
 
   config.save()
 
--- a/ufo/chrome_policy.py
+++ b/ufo/chrome_policy.py
@@ -25,7 +25,6 @@ def _make_chrome_policy_json():
   policy_dictionary = {
       "proxy_server_keys": proxy_server_public_keys,
       "enforce_proxy_server_validity": config.proxy_server_validity,
-      "enforce_network_jail": config.network_jail_until_google_auth,
   }
 
   return json.dumps(policy_dictionary)
@@ -69,6 +68,10 @@ def edit_policy_config():
   Returns:
     A redirect back to display chrome policy with will display the new values.
   """
+  # TODO(eholder): Move the display of config values and the edit handlers to
+  # something more sensible once UI review tells us what that should be. I'm
+  # envisioning a settings or options page which is underneath the overall
+  # Setup link. For now, I just want these settings to be edittable somewhere.
 
   config = ufo.get_user_config()
 
--- a/ufo/chrome_policy.py
+++ b/ufo/chrome_policy.py
@@ -34,11 +34,40 @@ def _MakeChromePolicyJson():
 @setup_required
 def chrome_policy():
   policy_json = _MakeChromePolicyJson()
+  config = models.Config.query.get(0)
 
-  return flask.render_template('chrome_policy.html', policy_json=policy_json)
+  return flask.render_template(
+      'chrome_policy.html', policy_json=policy_json,
+      enforce_proxy_server_validity=config.proxy_server_validity,
+      enforce_network_jail=config.network_jail_until_google_auth)
 
 
 @app.route('/chromepolicy/download/')
 @setup_required
 def chrome_policy_download():
   return _MakeChromePolicyJson()
+
+@app.route('/chromepolicy/edit', methods=['POST'])
+@setup_required
+def edit_policy_config():
+  """Post the form for editing the policy config values."""
+
+  config = models.Config.query.get(0)
+
+  # TODO(eholder): The Polymer html for sending toggle button values as inputs
+  # really sucks. I basically have to parse it out manually in JS then set it
+  # to a hidden input value. If we used iron-form, that may help with setting
+  # those as inputs on its own. That introduces a problem with refreshing the
+  # page on response or writing the response back into the UI for an update...
+  # TODO(eholder): There has to be a better way to convert strings to bools...
+  proxy_server_string = flask.request.form.get('enforce_proxy_server_validity')
+  network_jail_string = flask.request.form.get('enforce_network_jail')
+
+  config.proxy_server_validity = (proxy_server_string == 'true' or
+                                  proxy_server_string == 'True')
+  config.network_jail_until_google_auth = (network_jail_string == 'true' or
+                                           network_jail_string == 'True')
+
+  config.save()
+
+  return flask.redirect(flask.url_for('chrome_policy'))
--- a/ufo/chrome_policy.py
+++ b/ufo/chrome_policy.py
@@ -1,7 +1,5 @@
 """The module for generating and outputing chrome policy."""
 
-from . import app, setup_required
-
 import flask
 import json
 
@@ -30,8 +28,8 @@ def _make_chrome_policy_json():
   return json.dumps(policy_dictionary)
 
 
-@app.route('/chromepolicy/')
-@setup_required
+@ufo.app.route('/chromepolicy/')
+@ufo.setup_required
 def display_chrome_policy():
   """Renders the current chrome policy as json and editable values.
 
@@ -47,8 +45,8 @@ def display_chrome_policy():
       enforce_network_jail=config.network_jail_until_google_auth)
 
 
-@app.route('/chromepolicy/download/')
-@setup_required
+@ufo.app.route('/chromepolicy/download/')
+@ufo.setup_required
 def download_chrome_policy():
   """Outputs the managed chrome policy in json form for downloading as a file.
 
@@ -58,8 +56,8 @@ def download_chrome_policy():
   return flask.Response(_make_chrome_policy_json(),
                         mimetype='application/json')
 
-@app.route('/chromepolicy/edit', methods=['POST'])
-@setup_required
+@ufo.app.route('/chromepolicy/edit', methods=['POST'])
+@ufo.setup_required
 def edit_policy_config():
   """Receives the posted form for editing the policy config values.
 
--- a/ufo/chrome_policy.py
+++ b/ufo/chrome_policy.py
@@ -5,7 +5,8 @@ from . import app, setup_required
 import flask
 import json
 
-import ufo.models
+import ufo
+from ufo import models
 
 
 def _make_chrome_policy_json():
@@ -16,10 +17,10 @@ def _make_chrome_policy_json():
   Returns:
     A json string of current chrome policy.
   """
-  proxy_servers = ufo.models.ProxyServer.query.all()
-  proxy_server_public_keys = [s.make_public_key() for s in proxy_servers]
+  proxy_servers = models.ProxyServer.query.all()
+  proxy_server_public_keys = [s.get_public_key_as_authorization_file_string() for s in proxy_servers]
 
-  config = ufo.models.Config.query.get(0)
+  config = ufo.get_user_config()
 
   policy_dictionary = {
       "proxy_server_keys": proxy_server_public_keys,
@@ -39,7 +40,7 @@ def display_chrome_policy():
     The rendered chrom_policy.html template with policy values as variables.
   """
   policy_json = _make_chrome_policy_json()
-  config = ufo.models.Config.query.get(0)
+  config = ufo.get_user_config()
 
   return flask.render_template(
       'chrome_policy.html', policy_json=policy_json,
@@ -55,7 +56,8 @@ def download_chrome_policy():
   Returns:
     A json file of the current managed chrome policy.
   """
-  return _make_chrome_policy_json()
+  return flask.Response(_make_chrome_policy_json(),
+                        mimetype='application/json')
 
 @app.route('/chromepolicy/edit', methods=['POST'])
 @setup_required
@@ -68,7 +70,7 @@ def edit_policy_config():
     A redirect back to display chrome policy with will display the new values.
   """
 
-  config = ufo.models.Config.query.get(0)
+  config = ufo.get_user_config()
 
   # TODO(eholder): The Polymer html for sending toggle button values as inputs
   # really sucks. I basically have to parse it out manually in JS then set it
--- a/ufo/chrome_policy.py
+++ b/ufo/chrome_policy.py
@@ -9,7 +9,7 @@ import models
 
 
 def _MakeChromePolicyJson():
-  """Generate the json string of chrome policy based on values in the database.
+  """Generates the json string of chrome policy based on values in the db.
 
   This policy string has the following form:
 
@@ -32,7 +32,12 @@ def _MakeChromePolicyJson():
 
 @app.route('/chromepolicy/')
 @setup_required
-def chrome_policy():
+def display_chrome_policy():
+  """Renders the current chrome policy as json and editable values.
+
+  Returns:
+    The rendered chrom_policy.html template with policy values as variables.
+  """
   policy_json = _MakeChromePolicyJson()
   config = models.Config.query.get(0)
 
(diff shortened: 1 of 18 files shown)


xwsxethan:
chromemanagedpolicyui

Here's the UI so the reviewers can actually see what this looks like without pulling down the whole branch.

@jpevarnek PTAL as well.

ufo/chrome_policy.py
```diff
@@ -0,0 +1,73 @@
...
+
+from . import app, setup_required
+
+import flask
+import json
+
+import models
+
+
+def _MakeChromePolicyJson():
```
blueandgold: nit: think this might now be better as lowercase with underscore to be consistent with everything else here, and moving forward to our agreement to use lower casing

ufo/chrome_policy.py
```diff
@@ -0,0 +1,73 @@
...
+
+  return flask.render_template(
+      'chrome_policy.html', policy_json=policy_json,
+      enforce_proxy_server_validity=config.proxy_server_validity,
+      enforce_network_jail=config.network_jail_until_google_auth)
+
+
+@app.route('/chromepolicy/download/')
+@setup_required
+def chrome_policy_download():
```
blueandgold: nit: this method name should be better with the verb in front, download_chrome_policy()

ufo/chrome_policy.py
```diff
@@ -0,0 +1,73 @@
...
+      "enforce_proxy_server_validity": config.proxy_server_validity,
+      "enforce_network_jail": config.network_jail_until_google_auth,
+  }
+
+  return json.dumps(policy_dictionary)
+
+
+@app.route('/chromepolicy/')
+@setup_required
+def chrome_policy():
```
blueandgold: nit: only for the sake of having awesome code quality, please consider adding a verb to this method, e.g. display_chrome_policy(), or list_chrome_policy
otherwise, i have to read the code to get what this method is doing, which I would have gotten from the naming.

blueandgold: nit: docstring is needed?

blueandgold:
:+1:  Looks good to me, given that this is the initial implementation and it's something for John to take a look at.

xwsxethan:
Review status: 0 of 11 files reviewed at latest revision, 3 unresolved discussions.

---

*[ufo/chrome_policy.py, line 11 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9bxJYUq4valqHkpON5-r2-11:-K9hVjpIBVGjAxXnuIO1:-989169749) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/d1688c5f3fc3803eb1d2f585ff5b503177db8b4b/ufo/chrome_policy.py#L11)):*
We need to change just about all the other helper methods then, since all the ones except what is in ssh_client look like the previous method. Not saying we shouldn't do that, just note that we should check for this in future commits as well so that it gets cleaned up. Done.

---

*[ufo/chrome_policy.py, line 35 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9bxJYUq4valqHkpON5-r2-35:-K9hVzAoILKz83iK5FyA:2003197610) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/d1688c5f3fc3803eb1d2f585ff5b503177db8b4b/ufo/chrome_policy.py#L35)):*
Done and done.

---

*[ufo/chrome_policy.py, line 47 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9bxJYUq4valqHkpON5-r2-47:-K9hW8UfAmtCW8sZbZXN:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/d1688c5f3fc3803eb1d2f585ff5b503177db8b4b/ufo/chrome_policy.py#L47)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26)*

blueandgold:
Review status: 0 of 11 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[ufo/chrome_policy.py, line 11 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9bxJYUq4valqHkpON5-r2-11:-K9hWYgAgbs2cfoOWFgR:-837409305) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/d1688c5f3fc3803eb1d2f585ff5b503177db8b4b/ufo/chrome_policy.py#L11)):*
Yes, I know, that's why I mentioned "moving forward".  :)  At some point, we have to do a cleanup of our code.  Can we setup travis to enforce the casing and highlight what needs to "cleaned up"?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26)*

xwsxethan:
Review status: 0 of 11 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[ufo/chrome_policy.py, line 11 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9bxJYUq4valqHkpON5-r2-11:-K9hZrkjBEZN79pyukX5:-1012529579) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/d1688c5f3fc3803eb1d2f585ff5b503177db8b4b/ufo/chrome_policy.py#L11)):*
I think Landscape is (finally) back up and running, so I need to adjust our .pylintrc file to match with whatever we want. It sounds like we now want lower_case_with_underscores for all functions/methods right?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26)*

blueandgold:
Review status: 0 of 11 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[ufo/chrome_policy.py, line 11 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9bxJYUq4valqHkpON5-r2-11:-K9h_rOz02EsbRtEcF0D:-1213316713) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/d1688c5f3fc3803eb1d2f585ff5b503177db8b4b/ufo/chrome_policy.py#L11)):*
Yes, I think we all agreed to adopt the lowercase for method naming.  :)  Thanks for updating travis.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26)*

jpevarnek:
Reviewed 5 of 10 files at r1, 1 of 4 files at r2, 1 of 3 files at r3, 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[ufo/chrome_policy.py, line 22 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9mPGlsvSoaIWEDKckQ:-K9mPGlsvSoaIWEDKckR:-1429755830) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/chrome_policy.py#L22)):*
This should use `get_user_config`

---

*[ufo/chrome_policy.py, line 42 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9mPRmKjMb680F0cBVL:-K9mPRmKjMb680F0cBVM:951941073) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/chrome_policy.py#L42)):*
Is this used?

---

*[ufo/chrome_policy.py, line 58 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9mPatbt0vK7wXkySgt:-K9mPatbt0vK7wXkySgu:2092135451) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/chrome_policy.py#L58)):*
Set content type?

---

*[ufo/chrome_policy.py, line 78 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9mPjceH0o8vETgBe_g:-K9mPjceH0o8vETgBe_h:-1670271164) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/chrome_policy.py#L78)):*
That's because iron-form is meant to be used for asynchronous calls.  We've decided not to do that for simplicity.  If this is something we're doing often, create a new Polymer element (see my PR from earlier this week)

As far as the conversion, why are we sending them from the form as bools?!

---

*[ufo/models.py, line 135 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9m_vw-RGEKXoIkJmK9:-K9m_vw0Mc2w577qv_ba:-861779783) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/models.py#L135)):*
Can we rename this to something more descriptive?  This isn't making a public key so much as converting the binary representation of a public key that we already have to a representation that would fit in an authorized_keys folder.  Anyways, verb should definitely be get.

---

*[ufo/templates/chrome_policy.html, line 35 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9maKp3rLJaOZQkwOWJ:-K9maKp4OZSNRSO2QGVK:1329149641) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/templates/chrome_policy.html#L35)):*
Should use `on-tap` for this

---

*[ufo/templates/chrome_policy.html, line 55 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9mae1CcstF135NizVa:-K9mae1CcstF135NizVb:-359595245) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/templates/chrome_policy.html#L55)):*
Would something like `checked$=` work here?

---

*[ufo/templates/chrome_policy.html, line 75 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9may_pcxkhs1p-wX7D:-K9may_pcxkhs1p-wX7E:226137370) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/templates/chrome_policy.html#L75)):*
Should prefer `for (var i in toggles)` here

Also, can you explain why this is necessary here?  It seems like you're already inserting the values above.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26)*

xwsxethan:
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[ufo/chrome_policy.py, line 22 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9mPGlsvSoaIWEDKckQ:-K9mlUPxKqPEMzD9ZCK4:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/chrome_policy.py#L22)):*
Done.

---

*[ufo/chrome_policy.py, line 42 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9mPRmKjMb680F0cBVL:-K9mlVx0lvmnbHOWuVA6:1042683644) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/chrome_policy.py#L42)):*
Yes, it is used for config.proxy_server_validity and config.network_jail_until_google_auth which are passed to the template below.

I did switch it to get_user_config though.

---

*[ufo/chrome_policy.py, line 58 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9mPatbt0vK7wXkySgt:-K9mmQVpHVLYbBW5kLgV:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/chrome_policy.py#L58)):*
Done.

---

*[ufo/chrome_policy.py, line 78 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9mPjceH0o8vETgBe_g:-K9mmblohO-AN9YGsX3k:-221348436) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/chrome_policy.py#L78)):*
Yes, I know. That's why I was adding that into the comment since it isn't just a one-for-one exchange between standard forms and iron-forms, though iron-forms has better processing of paper and other Polymer elements, which aren't always handled correctly by standard forms.

The reason for using bools was since these are toggle switches that should only have two values. I was hoping to get something a little more analogous to a checkbox, just with a toggle UI overlaid. Our internal representation for these is bools, so it seemed to be a good fit. The check for 'true' or 'True' corresponds to 'true' set in javascript or 'True' returned from python and not altered in UI. I'm open to suggestions here as this part was unnecessarily messy. Standard html checkboxes would probably do the trick like a charm, but I was trying for something a bit nicer looking.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26)*

xwsxethan:
Was still writing comments before and accidentally published out of habit. Sorry!

---

Reviewed 5 of 10 files at r1, 1 of 4 files at r2, 1 of 3 files at r3, 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[ufo/models.py, line 135 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9m_vw-RGEKXoIkJmK9:-K9moDN203lIzqzdsnU1:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/models.py#L135)):*
Done.

---

*[ufo/templates/chrome_policy.html, line 35 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9maKp3rLJaOZQkwOWJ:-K9moJiIs4lvMtJFDnEK:-793234441) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/templates/chrome_policy.html#L35)):*
That didn't work... Click seems to work with mouse and keyboard, but not sure about mobile.

---

*[ufo/templates/chrome_policy.html, line 55 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9mae1CcstF135NizVa:-K9moKVpILAafWbwf4Zo:-1592228410) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/templates/chrome_policy.html#L55)):*
I tried that when initially writing this and couldn't get it to work. It didn't even register as the Polymer setting; it just added a separate attribute called checked$ which was entirely avoiding whenever I check or unchecked the toggle. Not sure if I did something wrong or if maybe it has to be encoded in a special way when passed from a jinja template.

As best I could tell, the support I'm looking for in the toggle button just doesn't exist without doing it yourself in javascript (at least as of this time).

---

*[ufo/templates/chrome_policy.html, line 75 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9may_pcxkhs1p-wX7D:-K9moyPGx7PCn4hsOeWn:-923456557) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/templates/chrome_policy.html#L75)):*
Done for the loop change.

So this is kind of janky to be honest, but basically there didn't seem to be a way to bind the state of the toggle to a value. In the if statement I have in jinja, I'm presetting checked (or not setting it) to match whatever state is returned. From there, I have to use this event listener to actually catch whenever the user toggles the switch. The event listener then copies that to the input element that will actually get sent with the form. The input element has the jinja variable it corresponds to as a name so that the backend form will process it correctly. None of this would be necessary if the paper-toggle-button allowed binding or was actually sent along with the form submit.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-:-K9mo8kK_nd6KhcNXitl:-1847277848)*

jpevarnek:
Reviewed 5 of 5 files at r5.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[ufo/chrome_policy.py, line 42 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9mPRmKjMb680F0cBVL:-KA1DIrGV4SZsXFcfHYC:-418320422) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/chrome_policy.py#L42)):*
Ah, sorry, missed that in everything below.  In that case, strong objection to including `enforce_network_jail` in Chrome policy, that would very definitely be a server-side item.

---

*[ufo/chrome_policy.py, line 78 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9mPjceH0o8vETgBe_g:-KA1DnKGj7Wclsqcjju9:1044230958) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/chrome_policy.py#L78)):*
Sorry, should probably have rephrased that.  What was the reasoning behind choosing to send the elements as the string representation of the boolean values?

---

*[ufo/templates/chrome_policy.html, line 35 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9maKp3rLJaOZQkwOWJ:-KA1ELQc91tyKu4fhyES:-1895793980) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/templates/chrome_policy.html#L35)):*
Sorry, that will only work inside of Polymer.  Any way to convert these areas to be structured with the business logic inside of Polymer elements?

---

*[ufo/templates/chrome_policy.html, line 55 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9mae1CcstF135NizVa:-KA1EkRvAoKLJhOqKQ4t:1434878427) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/templates/chrome_policy.html#L55)):*
Ah, sorry, missed that you were not putting this inside a real Polymer element.  See above comment.

---

*[ufo/templates/chrome_policy.html, line 75 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9may_pcxkhs1p-wX7D:-KA1GmiQltb15MyxXDHD:1330072656) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/templates/chrome_policy.html#L75)):*
Okay, any luck with building a custom Polymer element that handles the bindings?  If we are wanting to use Polymer, it would be nice to actually take advantage of things like this that it will handle for us.  Otherwise, there are plenty of much lighter-weight things we could use for the UI.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26)*

xwsxethan:
Review status: 8 of 15 files reviewed at latest revision, 7 unresolved discussions.

---

*[ufo/chrome_policy.py, line 42 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9mPRmKjMb680F0cBVL:-KA7-3GVuWU2DjouMaPI:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/chrome_policy.py#L42)):*
Done.

---

*[ufo/templates/chrome_policy.html, line 35 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9maKp3rLJaOZQkwOWJ:-KA77Nqfxv8TFHaUIas5:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/templates/chrome_policy.html#L35)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26)*

xwsxethan:
@jpevarnek I switched over to some custom polymer elements (based on your examples with the sidebar and textarea) which should greatly improve the logic here. Please take a look when you can. Thanks!

PS The changes to other templates beyond those related to chrome policy were to cleanup and use the new custom elements. Not directly related, but I wanted to roll the custom elements out across the board.

---

Reviewed 2 of 5 files at r5, 7 of 7 files at r6, 7 of 7 files at r7.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[ufo/chrome_policy.py, line 78 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9mPjceH0o8vETgBe_g:-KAB9FZHyiWseYkMOkgu:167585441) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/chrome_policy.py#L78)):*
I hadn't originally meant to use strings, but it became a necessity to convert from the jinja value into something that could be handled in javascript. However, in the new method, the presence (or lack) of checked as an attribute implicitly does this now on the client end and then simple json.loads converts it back on the server end. The new way should be much better.

---

*[ufo/templates/chrome_policy.html, line 55 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9mae1CcstF135NizVa:-KAB9oKAVs5YemBQysrU:-1726679816) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/templates/chrome_policy.html#L55)):*
Doing this in Polymer now so should be fixed. Please take a look.

---

*[ufo/templates/chrome_policy.html, line 75 \[r4\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-K9may_pcxkhs1p-wX7D:-KAB9uWIn6OHDvnrgusW:728895000) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/e2ad83d8050c5c8f4d98a228b9a12ba07252591e/ufo/templates/chrome_policy.html#L75)):*
Done as suggested. Thanks!

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-:-KABAYsnTkw2WO--HJ_e:-1384210664)*

jpevarnek:
Looks good now.  Still not thrilled with sending boolean values just as strings, but that's workable I guess.

---

Reviewed 3 of 7 files at r6, 5 of 7 files at r7, 2 of 2 files at r8.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/26#-:-KAHovaWSNmju9Cahsw5:1787200356)*

PR #30, opened 2016-02-11
Create Script to Vulcanize Polymer HTML into One File

xwsxethan:
To use this, simply run ./vulcanize.sh from the top level directory. It will install vulcanize for you, run bower install to make sure your packages are up to date, and then perform the vulcanization.

The vulcanization basically works by finding every html file under ufo/static/. It then creates a temporary input html file and writes each found file to the input as an import, such as:
`<link rel="import" href="foundFile01.html" />`
`<link rel="import" href="foundFile02.html" />`
`<link rel="import" href="foundFile03.html" />`
...
`<link rel="import" href="foundFileN.html" />`

Afterwards, it runs vulcanize on this input file to generate one master output file which contains all Polymer elements currently installed as well as the custom elements listed under static (such as sidebar.html and textarea.html).

I removed all references in our templates and added a reference to this output file in layout.html so that each template which inherits from layout will pick this up.

NOTE: After doing this, it seems that the font has changed on some pages. This is likely because I pulled in a font that we hadn't specifically referenced yet. It doesn't look bad by any means, but it just a difference you may notice after doing this.

Moving forward, if we install a new bower package or create a new custom element, you will need to rerun vulcanize.sh to generate a new output file. However, the script should handle this all for you automatically (with the exception of prompting for your password when sudo installing vulcanize).

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/30)

### DIFF
--- a/ufo/templates/layout.html
+++ b/ufo/templates/layout.html
@@ -2,9 +2,7 @@
 <html lang="en">
 <head>
   <title>{% block title %}{% endblock %}</title>
-  <link rel="import" href="{{ url_for('static', filename='bower_components/polymer/polymer.html') }}" />
-  <link rel="import" href="{{ url_for('static', filename='bower_components/paper-drawer-panel/paper-drawer-panel.html') }}" />
-  <link rel="import" href="{{ url_for('static', filename='sidebar.html') }}" />
+  <link rel="import" href="{{ url_for('static', filename='vulcanized.html') }}" />
   <link rel="stylesheet" href="{{ url_for('static', filename='style.css') }}" />
   {% block head %}{% endblock %}
 </head>
--- a/vulcanize.sh
+++ b/vulcanize.sh
@@ -95,9 +95,10 @@ function vulcanizeSingleFileForImports ()
 function printHelp ()
 {
   echo
-  echo "Usage: vulcanize.sh [help|--help]"
+  echo "Usage: vulcanize.sh [help|--help|travis]"
   echo
   echo "  help, --help   - Prints this help text."
+  echo "  travis         - Creates the vulcanized file in Travis CI."
   echo
   echo "If no parameters are specified, this generates a vulcanized html file "
   echo "of bower packages and custom elements which were found under "
@@ -111,6 +112,12 @@ function printHelp ()
 
 if [ "$#" == 0 ] || [ ! "$1" == 'help' && ! "$1" == '--help' ]; then
   installVulcanize
+  updateBowerPackages
+  findHtmlFilesToVulcanize
+  createSingleHtmlFileToVulcanize
+  vulcanizeSingleFileForImports
+elif [ "$1" == 'travis' ]; then
+  runAndAssertCmd "npm install -g vulcanize"
   updateBowerPackages
   findHtmlFilesToVulcanize
   createSingleHtmlFileToVulcanize
--- a/vulcanize.sh
+++ b/vulcanize.sh
@@ -57,12 +57,25 @@ function findHtmlFilesToVulcanize ()
   runInStaticDirAndAssertCmd "rm -fr $TEMP_FILE_LIST"
   runInStaticDirAndAssertCmd "rm -fr $HTML_FILE_TO_VULCANIZE"
   runInStaticDirAndAssertCmd "touch $TEMP_FILE_LIST"
+  # This searches through the ufo/static/ directory and finds all files that
+  # match *html, but excludes index.html, basic.html, files under directories
+  # with test in the name, and files under directories with demo in the name.
+  # It then sed's the files to remove the prepended ./ and outputs to
+  # $TEMP_FILE_LIST.
   runInStaticDirAndAssertCmd "find . -name '*html' -not -name 'index.html' -not -name 'basic.html' -not -path '*test*' -not -path '*demo*' | sed 's|./||' > $TEMP_FILE_LIST"
 }
 
 function createSingleHtmlFileToVulcanize ()
 {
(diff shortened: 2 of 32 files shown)


xwsxethan:
@jpevarnek to review as well.

xwsxethan:
vulcanizationscreenshot

Here is a screenshot of the new requests going across the wire. Highlighted in blue are all fonts. As you can see, outside of the fonts, the only requests are for the template itself, the vulcanized html file, and our specific css style and custom js. The last two could be added to the vulcanized html file as well, but I left them out for now so as not to require re-vulcanizing every time we make a change to either of those.

ufo/templates/layout.html
```diff
@@ -2,9 +2,7 @@
 <html lang="en">
 <head>
   <title>{% block title %}{% endblock %}</title>
-  <link rel="import" href="{{ url_for('static', filename='bower_components/polymer/polymer.html') }}" />
-  <link rel="import" href="{{ url_for('static', filename='bower_components/paper-drawer-panel/paper-drawer-panel.html') }}" />
-  <link rel="import" href="{{ url_for('static', filename='sidebar.html') }}" />
+  <link rel="import" href="{{ url_for('static', filename='vulcanized.html') }}" />
```
blueandgold: Just to double-check here, vulcanized.html will contain every polymer element that we are using in the app, and even pages that don't use certain elements will load them all?  It's better than what we have before (all those separate little requests).

vulcanize.sh
```diff
@@ -0,0 +1,104 @@
...
+  runInStaticDirAndAssertCmd "rm -fr $TEMP_FILE_LIST"
+  runInStaticDirAndAssertCmd "rm -fr $HTML_FILE_TO_VULCANIZE"
+  runInStaticDirAndAssertCmd "touch $TEMP_FILE_LIST"
+  runInStaticDirAndAssertCmd "find . -name '*html' -not -name 'index.html' -not -name 'basic.html' -not -path '*test*' -not -path '*demo*' | sed 's|./||' > $TEMP_FILE_LIST"
+}
+
+function createSingleHtmlFileToVulcanize ()
+{
+  runInStaticDirAndAssertCmd "touch $HTML_FILE_TO_VULCANIZE"
+  runInStaticDirAndAssertCmd 'while read line; do echo "${IMPORT_BEFORE_STATEMENT}${line}${IMPORT_AFTER_STATEMENT}"; done < $TEMP_FILE_LIST > $HTML_FILE_TO_VULCANIZE'
```
blueandgold: My bash-fu is not that strong, but what exactly is happening here?  I can't tell how it knows what to vulcanize from $TEMP_FILE_LIST.  Can we add some documentation here to explain?

vulcanize.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+
+# Script to vulcanize all the Polymer files into one big blob.
```
blueandgold: I take it that this script is not hooked into our build process?  Is there a way to do so?

blueandgold:
Looks great overall!  I like the approach to break down the problem for the polymer elements first, instead of the whole page.  Just some questions to clarify my understanding.

ufo/templates/layout.html
xwsxethan: Yes, it contains every custom element defined in static (sidebar.html, textarea.html, and any future custom ones we defined, such as toggleInput or collapseButton from my other PR's) and every bower_component currently downloaded. This will obviously be way more than any single page needs individually, but the tradeoff is that this file is cached after initial load which means after initial load, there is no need to redownload the vulcanized file. Navigating from the user handler to the proxy server handler for instance will then just pull down the requested template (if modified) and surface the cached vulcanized file.

For the hard data, vulcanized.html is (currently) 731.3 kB and downloading an updated version (I added a space and refreshed the page) took anywhere between 14 and 30 ms on my machine. When it is cached, the size of the total request to re-surface the cached version is only 415 B which takes anywhere from 8 ms to 27 ms on my machine. If I jump back on master without any of these changes though, loading any one element is roughly 412 B from cache and takes somewhere between 3 and 40 ms. There are roughly 30~40 elements per page though, as one element may depend on many more recursively.

vulcanize.sh
xwsxethan: I'll add some comments to explain, but here it is in a nutshell:

We find every html file and format it to look like this: `/bower_components/component_name_here/some_component.html`

Instead of passing that list around, I write it to $TEMP_FILE_LIST with one of those html file references per line. In line 66, I then read  $TEMP_FILE_LIST back, line by line, and add the pieces to each line to make it into an import, such as:
`<link rel="import" href="/bower_components/component_name_here/some_component.html" />`

The import statements get written out to another file, $HTML_FILE_TO_VULCANIZE. The reason for outputting a file and reading it back was simply for debugging purposes. I could just sed it all on the fly instead, but this allowed me to have some record of what it failed on in case we encounter errors. From there, $HTML_FILE_TO_VULCANIZE with all the import statements gets vulcanized into the final output. Both of the temp files are deleted on success, so they don't stick around unless something fails.

vulcanize.sh
xwsxethan: The tests were passing before, so perhaps it isn't needed?? I just added it anyway.

vulcanize.sh
blueandgold: Okay, that helps.  But I'm scratching my head of how each `/bower_components/component_name_here/some_component.html` are being found, especially since they are deleted from each html now?

blueandgold:
:+1:   Thanks for explaining to me that the polymer components are gotten from a local bower directory under ufo/static.  Please just add a comment so that we will remember this.  :)

PR #31, opened 2016-02-11
Cleaning up some miscelaneous problems with our HTML

xwsxethan:
Such as 'hae' should be 'have' in the setup page, and adding a br tag and div to the proxy server page so that the top button and card don't overlap, like in the user page.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/31)


blueandgold:
Cool!  :+1:

PR #32, opened 2016-02-11
Making the Invite Code into a URL with uProxy Prefixed

xwsxethan:
Also adding a button to directly access the link if desired.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/32)

### DIFF
--- a/ufo/user.py
+++ b/ufo/user.py
@@ -14,7 +14,7 @@ from ufo import models
 from ufo import oauth
 from ufo import setup_required
 
-INVITE_CODE_URL_PREFIX = 'https://uproxy.org/connect/'
+INVITE_CODE_URL_PREFIX = 'https://uproxy.org/connect/#'
 
 def _render_user_add(get_all, group_key, user_key):
   credentials = oauth.getSavedCredentials()
--- a/ufo/user.py
+++ b/ufo/user.py
@@ -143,10 +143,11 @@ def user_details(user_id):
   user = models.User.query.get_or_404(user_id)
   invite_code = _MakeInviteCode(user)
   if invite_code is None:
-    invite_code = ''
-  invite_url = INVITE_CODE_URL_PREFIX + invite_code
-  return flask.render_template('user_details.html', user=user,
-                               invite_url=invite_url)
+    return flask.render_template('user_details.html', user=user)
+  else:
+    invite_url = INVITE_CODE_URL_PREFIX + invite_code
+    return flask.render_template('user_details.html', user=user,
+                                 invite_url=invite_url)
 
 @app.route('/user/<user_id>/delete', methods=['POST'])
 @setup_required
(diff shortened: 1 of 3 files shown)


xwsxethan:
@blueandgold to review as well.

ufo/user.py
```diff
@@ -139,9 +144,13 @@ def add_user():
...
 def user_details(user_id):
   user = models.User.query.get_or_404(user_id)
-  return flask.render_template('user_details.html',
-                               user=user,
-                               invite_code=_MakeInviteCode(user))
+  invite_code = _make_invite_code(user)
+  if invite_code is None:
+    return flask.render_template('user_details.html', user=user)
+  else:
+    invite_url = INVITE_CODE_URL_PREFIX + invite_code
```
blueandgold: Do we still need the # symbol to secure the invite code in the URL as per the security review?

xwsxethan: Oops, yeah that was supposed to be in the prefix constant, but I think I accidentally undo-ed one too many times and removed it. Added back now.

blueandgold:
Awesome, looks good to me. :+1:

jpevarnek:
Reviewed 2 of 3 files at r1, 1 of 1 files at r2.

---

*[ufo/user.py, line 17 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/32#-KAHrCRd0f2wLv_h6smv:-KAHrCRd0f2wLv_h6smw:-504861320) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/6dc954bee5c165d21b473dbd1f737e7a95ce889a/ufo/user.py#L17)):*
I think it would look better with that last forwards lash dropped, your call completely though.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/32)*

xwsxethan:

---

*[ufo/user.py, line 17 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/32#-KAHrCRd0f2wLv_h6smv:-KALAPUaok047h-8r829:-1167409497) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/6dc954bee5c165d21b473dbd1f737e7a95ce889a/ufo/user.py#L17)):*
Will that work without the forward slash? I'm checking here (https://github.com/uProxy/uproxy/blob/6f665fb90ef3884c3543b655d98e6e1507cd4f21/src/generic_ui/scripts/ui.ts#L558) at the parseInvite function, which seems to be based on last index of forward slash in the string (unless this needs to be updated as well, which is fine if so). I am fine with either and happy to change it, just want to make sure it matches and works as intended.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/32)*

PR #36, opened 2016-02-18
Polymer Client-Side Validation for Free-form Inputs

xwsxethan:
This was initially a change for the old UI, but I added support in the new UI now as well. We're checking email addresses, public keys, private keys, ip addresses, and user/group keys during lookup (which can be an email address or number). I also tweated the UI slightly so that input elements have more spacing around them to allow for the error text if it should appear.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/36)


blueandgold:
Sorry, I don't know what happened, but I didn't see this PR until now.  :(  Looks great overall, just a couple of things.

---

Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion.

---

*[ufo/__init__.py, line 37 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/36#-KC1iQvZE0RN908NDrzr:-KC1iQvZE0RN908NDrzs:292212342) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/60ff13a4511af7a2232b45469f7e367aba18b098/ufo/__init__.py#L37)):*
Everything looks great overall.  Just a couple of questions:

#1: Now that we are not using jinja_env.globals, do you think that this (__init__.py) is still the right place for these regex patterns?  How about creating a regex.py and put that in the services package?  This would also reduce the bloat here in __init__.py.

#2: Can we add some unit test to make sure all the patterns works as intended?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/36#-:-KC1jRWp1UBzsImeFjiy:-1745865121)*

xwsxethan:
Reviewed 8 of 8 files at r1, 5 of 5 files at r2.

---

*[ufo/__init__.py, line 37 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/36#-KC1iQvZE0RN908NDrzr:-KC2DIrtMkjNkNxjvuk6:-340666474) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/60ff13a4511af7a2232b45469f7e367aba18b098/ufo/__init__.py#L37)):*
Done and done. Thanks for the suggestion!

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/36)*

blueandgold:
Sweet!  :+1:

PR #37, opened 2016-02-19
Save user key file on the proxy server.

blueandgold:
The goal of this PR is to save the user keys on the proxy servers.
- We currently do not know which file exactly to save the user keys to.  Jonathan will follow-up, so saving it temporarily, with a TODO.
- Per DD, we will need to notify the admin if the key distribution fails.  Since we have not figure out what's the best way to do this, this is another TODO.
- I would appreciate some suggestion on testing this.  Since we want to avoid writing unit tests that are coupled to the app code, I am having a hard time to see what useful unit test would be worth writing here.  Instead, I think that we almost need to write a monitoring job that will functionally check this is working correctly. 
- Added exception handling in the ssh client if ssh connection can not be established.
- Since I am not sure what file we are writing to, I just decided to write the user keys to a brand new file on each update.  It's pretty easy to change if we need to concat it instead.
- This is a branch off `jobqueue > logger`, and will need to be submitted after those.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/37)

### DIFF
--- a/ufo/key_distributor.py
+++ b/ufo/key_distributor.py
@@ -50,7 +50,7 @@ class KeyDistributor(object):
 
   def enqueue_key_distribution_jobs(self):
     """Distribute user keys to proxy servers to authenticate invite code."""
-    key_string = self._make_key_string()
+    key_string = self.make_key_string()
     proxy_servers = models.ProxyServer.query.all()
     queue = Queue(connection=worker.CONN)
     for proxy_server in proxy_servers:
--- a/bower.json
+++ b/bower.json
@@ -8,7 +8,7 @@
     "polymer": "~1.2.3",
     "paper-button": "PolymerElements/paper-button#~1.0.11",
     "paper-card": "PolymerElements/paper-card#~1.0.8",
-    "paper-checkbox": "PolymerElements/paper-checkbox#~1.0.16",
+    "paper-checkbox": "PolymerElements/paper-checkbox#~1.1.0",
     "paper-drawer-panel": "PolymerElements/paper-drawer-panel#~1.0.6",
     "paper-input": "PolymerElements/paper-input#~1.1.3",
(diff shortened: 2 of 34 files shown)


blueandgold:
@jpevarnek PTAL as well, thanks!

ufo/key_distributor.py
```diff
@@ -44,9 +44,34 @@ def _distribute_key(self, proxy_server, key_string):
...
     client = ssh_client.SSHClient()
-    client.connect(proxy_server)
 
-    # TODO do stuff
+    try:
+      client.connect(proxy_server)
+    except (ssh_client.SSHConnectionException, 
+            ssh_client.InvalidKeyTypeException):
+      app.logger.error(
+          'Unable to connect to proxy server to distribute keys: %s',
```
xwsxethan: Maybe move the %s string by the proxy server so it doesn't look like we're writing the keys out to the logs. Obviously, you are just writing the proxy server's name, but it reads kind of funny here.

xwsxethan:
This looks fine to me overall. You may want to pull out the file name to a constant so it's easier to find later, but that's about the smallest nit I could pick. I really like the exception handling.

Per your question about testing, I think monitoring makes sense here as well. Whatever time interval we have between syncs, X, if a server's key file has not been touched in say 2X, then the test has failed. You could force some exceptions to be thrown in a unit test and assert that they are handled as we expect, but that's a stretch. I'll leave that up to you. :+1: for me (pending the prior PR of course).

ufo/key_distributor.py
blueandgold: Done.  Reads much nicer now.

blueandgold:
Thanks for the review!  Both of your comments are addressed.

PR #42, opened 2016-02-22
Fix Vulcanization on Heroku by Adding to the Postdeploy Process in App.json

xwsxethan:
The vulcanization script won't create the necessary file when run after deploying to heroku so I've made it part of the deployment in app.json. This also adds vulcanization to the dependencies in package.json so that it is already available without having to be installed manually.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/42)

### DIFF
--- a/app.json
+++ b/app.json
@@ -7,7 +7,7 @@
   "logo": "https://uproxy.org/images/uproxy_logo.svg",
   "success_url": "/setup",
   "scripts": {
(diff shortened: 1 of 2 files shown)


blueandgold:
Very nice!

blueandgold:
:+1:

PR #43, opened 2016-02-23
Fixing Vulcanize on Heroku (Again)

xwsxethan:
Moving from postdeploy event step in app.json to postinstall event step in package.json which seemed to work in another branch.

My previous PR didn't actually address the issue, but my changes to allow the vulcanized.html file in gitignore (and subsequent changes to restore gitignore) masked the issue. Now it is generating the correct html file as part of the build without us checking in vulcanized.html to source control. If there are any issues using this, please let me know ASAP, and I'll investigate.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/43)

### DIFF
--- a/app.json
+++ b/app.json
@@ -7,7 +7,7 @@
   "logo": "https://uproxy.org/images/uproxy_logo.svg",
   "success_url": "/setup",
   "scripts": {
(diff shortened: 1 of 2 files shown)


blueandgold:
Thanks for giving this another go!  :+1:

PR #46, opened 2016-02-25
Overlay UI

xwsxethan:
I'm just starting this PR so I can attach screenshots as necessary. I'm still working on the dialog/overlay, but so far so good. I've also fixed a few other minor things, listed below:
- Moved image assets to static/img just for cleaning up that folder.
- Moved the body section from the base template out of the header panel. This allows us to more easily stack elements as well as reposition the body so that it is on top of the toolbar (as specified in the mocks).

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/46)


xwsxethan:
See screenshots below of the moved top paper-material element (on top of the toolbar at the top) and the new overlay (very basic so far):
ufonewui07
ufonewui08

xwsxethan:
Pics of new flow:
ufoadduserui01
ufoadduserui02
ufoadduserui03

xwsxethan:
I still have to finish up a few things (see the comment on my last commit), but I figured you could go ahead now since this is mostly done @blueandgold .

bower.json
```diff
@@ -5,12 +5,14 @@
     "iron-ajax": "PolymerElements/iron-ajax#~1.1.1",
     "iron-collapse": "PolymerElements/iron-collapse#~1.0.4",
     "iron-flex-layout": "PolymerElements/iron-flex-layout#~1.2.3",
-    "iron-form": "PolymerElements/iron-form#~1.0.13",
+    "iron-form": "PolymerElements/iron-form#~1.0.14",
```
blueandgold: nit: Just being curious, is there some easier or more automated way to keep the versions updated?  Just asking and not something that needs to be addressed here.

ufo/static/userAddForm.html
```diff
@@ -0,0 +1,219 @@
...
+        <paper-spinner active$={{loading}}></paper-spinner>
+      </template>
+      <iron-ajax
+        url="{{resources.addUrl}}"
+        method="POST"
+        handle-as="json"
+        id="postItemsAjax"
+        loading="{{loading}}"
+        last-response="{{ajaxResponse}}"
+        on-response="bubbleToList">
```
blueandgold: nit: can we make the naming of "bubbleToList" more clear?  It's hard to give an example as it's not obvious to me what this is doing, even when I look at the function itself.  :)

blueandgold:
+1 Looks good to me.  Please watch out for merging into the new packages.

xwsxethan:
Reviewed 12 of 19 files at r1, 6 of 14 files at r2, 4 of 6 files at r3, 3 of 6 files at r4, 2 of 2 files at r5.

---

*[bower.json, line 8 \[r3\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/46#-KBOBW7AizaSj3iDMKW5-r3-8:-KBhRjJ_zijAftF4Hh5Q:1135785399) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/daa2762d6f33a442d86a3d74f2bebdbf756614db/bower.json#L8)):*
This probably didn't need to be updated, but I just went ahead with it as I was developing to ensure that the docs I was reading at the time were in sync with the current version (for my own sanity as I was having some issues using forms at the time).

As far as keeping the versions updated, I don't know that we'll need to change these very often unless we want to update past some minor change. Here's a better explanation than I can give on the versioning syntax for bower and what it does for us:
http://stackoverflow.com/questions/19030170/what-is-the-bower-version-syntax

In brief, bower will never give us a package that would break us, but it will keep us up to date with patches. We could modify this to update on minor changes, but nothing that breaks our API if we choose. The problem would be that Polymer components may not break API-wise, but their aesthetic could change, leading to poor UX.

---

*[ufo/static/userAddForm.html, line 123 \[r3\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/46#-KBTP0xpL-sj8oxqGhVF-r3-123:-KBhTixR8ofe0XTgMuLO:276781415) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/daa2762d6f33a442d86a3d74f2bebdbf756614db/ufo/static/userAddForm.html#L123)):*
Changed to sendUsersJsonToList as that's more descriptive. Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/46)*

ufo/static/addUserModal.html
```diff
@@ -0,0 +1,132 @@
...
+                  {{resources.dismissText}}
+                  </paper-button>
+                  <paper-button class="anchor-button" autofocus on-tap="submitPostForm">
+                  {{item.saveButton}}
+                  </paper-button>
+                </div>
+              </div>
+            </template>
+            <template is="dom-if" if="{{!item.isManual}}">
+              <user-add-form resources="{{resources}}" add-flow-text-obj="{{item}}"></user-add-form>
```
blueandgold: It looks like this form can be re-used in the setup page to add users as well.  Do you expect this custom element to be just dropped in there (i.e. work outside of the modal window)?  It looks okay to me, but want to verify this.

xwsxethan:

---

*[ufo/static/addUserModal.html, line 76 \[r5\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/46#-KBOZb2eWBUqVqBvo250-r5-76:-KBiSdVYAGiuJvYX0_zI:2140907284) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/40ce7011e545a117eab4af9a1b22e71919ea7aa2/ufo/static/addUserModal.html#L76)):*
Sort of, yes. I think we'll actually be able to reuse almost all of this as you intend, but it won't be confined to the modal. I just emailed you about something similar to this, but I didn't realize you needed the modal as well.

Let me send off a separate PR once I've merged this portion, if you are okay with this work that is?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/46)*

blueandgold:

---

*[ufo/static/addUserModal.html, line 76 \[r5\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/46#-KBOZb2eWBUqVqBvo250-r5-76:-KBiUrVx6fin8g_lileG:-1808562452) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/40ce7011e545a117eab4af9a1b22e71919ea7aa2/ufo/static/addUserModal.html#L76)):*
Closing the loop here.   We had an offline chat, and to be clear I need the form inside the modal, and not the modal itself.  The screenshot below will make everything clear.  

https://screenshot.googleplex.com/fnMJvZToRz8

As can be seen, we need to de-couple the form from modal so that the form can be reused elsewhere.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/46)*

PR #53, opened 2016-03-01
Add proxy server form to the setup page

blueandgold:
There is a bit of work to make the proxy server form to, well, work.
- I changed the xsrf token to be per-session basis, instead of per-request basis.  This is a simple solution to the problem of folks adding multiple servers, one after another, on the setup page.  In which case, the token would have to be refreshed, and updated on all the forms of the setup page.  On a high-level, [per-session xsrf token actually is okay](http://security.stackexchange.com/questions/22903/why-refresh-csrf-token-per-form-request).
- Resetting the form upon response, instead of closing the modal window.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/53)

### DIFF
--- a/ufo/handlers/setup.py
+++ b/ufo/handlers/setup.py
@@ -7,8 +7,9 @@ import oauth2client
 
 import ufo
 from ufo.handlers import chrome_policy
-from ufo.services import oauth
 from ufo.handlers import user
+from ufo.handlers import proxy_server
+from ufo.services import oauth
 
 
 DOMAIN_INVALID_TEXT = 'Credentials for another domain.'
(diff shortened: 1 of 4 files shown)


xwsxethan:
The reviewable link isn't showing up, but overall this looks good to me. Did you verify that the (x/c)srf token is still different for different sessions (just to make sure this works as intended)? If that works, then it would be awesome.

I'll probably steal the approach here and followup to make sure the add form follows the same basic idea for resetting.
:+1:

blueandgold:
Good thinking.  I double-checked just now and the xsrf token is indeed reset on a new session.

PR #54, opened 2016-03-01
Fixing Form and Modal Reset on the Landing and Setup Pages

xwsxethan:
This should now work correctly with multiple subsequent posts.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/54)

### DIFF
--- a/ufo/static/userAddForm.html
+++ b/ufo/static/userAddForm.html
@@ -42,13 +42,13 @@
             <paper-listbox>
               <template is="dom-repeat" items="[[lastResponse.directory_users]]">
                 <paper-item id="{{item.name}}" class="horizontal layout">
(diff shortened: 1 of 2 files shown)


ufo/static/userAddForm.html
```diff
@@ -113,6 +113,7 @@
         this.querySelector('#' + this.addFlowTextObj.id).submit();
       },
       submitPostForm: function() {
+        this.setUsersFromCheckboxes();
```
blueandgold: nit: When I read this line here, I feel I need more context on what user we are setting.  So, how about the name below?
```javascript
this.setSelectedUsersFromCheckboxes();
```

blueandgold:
Very nice, just one small nit. :+1:

PR #57, opened 2016-03-03
Fix the body from bleeding into the header.

blueandgold:
The setup page had a problem where the body was bleeding into the header.

The fix was simply to surround the paper-scroll-header-panel  with `<div></div>` tag.
(99% of the work was in troubleshooting.)

What the setup page looks like now.  I haved checked the new landing page to make sure this doesn't negatively affect it.
screenshot from 2016-03-03 13 43 07

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/57)

### DIFF
--- a/ufo/static/style.css
+++ b/ufo/static/style.css
@@ -146,7 +146,7 @@ paper-icon-item {
   /* This is used to make the body overlap a bit with the header.
      But, it is commented out because the body tends to bleed more and
      more into the header as the body gets longer and longer.
-     So, we need to find a more robust way to do this.
+     TODO: We need to find a more robust way to do this.
   margin-top: -80px;
   */
 }
--- a/ufo/static/style.css
+++ b/ufo/static/style.css
@@ -143,5 +143,10 @@ paper-icon-item {
 }
 
 #main-holder {
+  /* This is used to make the body overlap a bit with the header.
+     But, it is commented out because the body tends to bleed more and
+     more into the header as the body gets longer and longer.
+     So, we need to find a more robust way to do this.
   margin-top: -80px;
+  */
 }


xwsxethan:
Ahhh so this was somewhat by intent. Obviously, the overlap needs to be perfected, but per the mocks, the cards should overlap the header slightly. The slightly part is tricky as the cards seem to want to position higher and higher as more users are added. I haven't figured out how to do that yet...

However, you shouldn't add that div there. If you really want to disable this for now, take the div back out and go to style.css. At the bottom is an id tag for #main-holder where the margin-top is set to some negative number. Just comment that out or remove it and leave a TODO to come back and fix it. That's where it is actually being set currently.

For the fix later, there is probably some way to use calc to figure out how much the main-holder div needs to be adjusted negative in order for the overlap to work correctly based on the current number of users.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/57#-:-KByVxj-Me5AyxzgqDHk:-153695276)*

blueandgold:
Done, commenting out the offending css rule.  Also undoing the previous commit.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/57#-:-KByZtwCgKQykzZfIdia:-817003184)*

xwsxethan:
:lgtm_strong: Thank you!

---

Reviewed 1 of 2 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/57#-:-KBy_QwgE_die4jiBjYh:-1913360213)*

PR #58, opened 2016-03-03
User Hover and Details UI

xwsxethan:
This PR implements the hover and dialog overlay for details functionality on users on the landing page. I had to change how we get invite codes slightly to make this more usable with AJAX, but the format of the invite code itself remains identical to before. I also stubbed out a custom element for the edit server dialog, which should be more straightforward to implement in a separate PR. I will follow up with that when the time comes. Please see the attached screenshots for more info.

[Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/58)

ufouserdetailsui01
ufouserdetailsui02


ufo/handlers/user.py
```diff
@@ -261,63 +274,79 @@ def user_details(user_id):
...
   error handler.
 
-  Args:
-    user_id: A string identifying a user in the database.
-
   Returns:
     A redirect to the user_list page after deleting the given user.
   """
+  json_id = flask.request.form.get('user_id')
+  user_id = json.loads(json_id)
```
blueandgold: nit: It's actually my fault to be picky here, but I think these 2 lines would actually read better if they are folded into one line.  i.e. toss the "json_id"
```python
user_id = json.loads(flask.request.form.get('user_id'))
```

ufo/handlers/user.py
```diff
@@ -261,63 +274,79 @@ def user_details(user_id):
...
   error handler.
 
-  Args:
-    user_id: A string identifying a user in the database.
-
   Returns:
     A redirect to the user_details page after rotating the user's keys.
   """
+  json_id = flask.request.form.get('user_id')
+  user_id = json.loads(json_id)
```
blueandgold: ditto

ufo/handlers/user.py
```diff
@@ -261,63 +274,79 @@ def user_details(user_id):
...
   If the user is not found, this produces a 404 error which redirects to the
   error handler.
 
-  Args:
-    user_id: A string identifying a user in the database.
+  Returns:
+    A json object with 'invite_code' set to the invite code string.
+  """
+  json_id = flask.request.args.get('user_id')
+  user_id = json.loads(json_id)
```
blueandgold: ditto (everywhere)

blueandgold:
:+1:   Nice!

xwsxethan:
Done across the board. Thanks! Will merge once the tests complete.

PR #60, opened 2016-03-05
Extract the sectional header into a separate custom element.

blueandgold:
- This a very rough PR to test the general approach.
- The problem I'm trying to solve here is that sectional header (I don't know what to call it exactly) where the title text is being displayed is not necessary for the setup page. e.g. add user section has just a large add user icon on the left-side.  Please see the UI mock or the attached screenshot.
- The simplest and the cleanest way I can see to address this is to extract the header portion into a separate custom element, instead of adding more conditional templating inside paperDisplayTemplate.  This way, it's easy for someone to just apply the header element if they need it.
- What do you think of this general approach?  I am open to any other ways to solve this.  (Please don't mind the UI styling which still needs to be tightened.)
- I tested the new landing page with this change, and it works fine.

screenshot from 2016-03-04 16 15 01

---

This change is [Review on Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/60)

### DIFF
--- a/ufo/static/headerContainer.html
+++ b/ufo/static/headerContainer.html
@@ -1,0 +1,37 @@
+<link rel="import" href="bower_components/iron-ajax/iron-ajax.html" />
+<link rel="import" href="bower_components/paper-material/paper-material.html" />
+<link rel="import" href="bower_components/paper-toolbar/paper-toolbar.html" />
+
+<dom-module id="header-container">
+  <template>
+    <div class="heading">
+      <paper-toolbar class="medium-tall">
+        <div class="middle flex">
+          <div class="horizontal center-justified layout">
+            <div class="title">{{resources.titleText}}</div>
+            <template is="dom-if" if="{{resources.hasAddFlow}}">
+              <add-item-button resources="{{resources}}"></add-item-button>
+            </template>
+          </div>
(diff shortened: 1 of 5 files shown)


xwsxethan:
This looks basically good to me, but it seems like you're still using paper-display-template with the new header everywhere. Did you mean to not use the template anymore on the setup page in this PR (or is that going to be separate) or am I missing something here? Otherwise, :+1: from me.

blueandgold:
Correct, I am still using the paper-display-template, as it serves to position all the components the page.  So, it is good as is.  

To clarify, what I was not sure is: should we continue to add to the conditional templating by using another variable to control whether to display the header text, vs a custom element as a wrapper and save a variable.

This PR was to test the latter approach with the wrapper element, which seems nicer by breaking things down a bit.  Further thought?

xwsxethan:
That makes sense, and this approach is fine. I was missing the part in setup where you haven't applied the header (I didn't expand the diff, so it looked like the header was still applied everywhere). Sounds good to me though!

blueandgold:
Cool, let's see how this works, and we can rework it when we get a better sense of how to structure the UI and/or polymer.

PR #63, opened 2016-03-15
Inverting the color scheme on the user and server add svg files

xwsxethan:
This lets the icons actually show up in the button since now the icon is white with a clear background instead of green with the clear background. Also added some space between the text and icon for clarity.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/63)


blueandgold:
Cool!  :+1:

PR #69, opened 2016-03-15
Pull Text from the OAuth Form Out of HTML and Into JSON

xwsxethan:
Refactoring the text on the OAuth form to be specified from the server via JSON, rather than hardcoded into the html page. This will help out for i18n later. I was doing this in another PR and decided to move this out since it wasn't really related.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/69)


blueandgold:
Cool..!! :+1:

PR #71, opened 2016-03-16
Server Edit Dialog UI

xwsxethan:
From talking with John Cassidy, the preferred UI flow here would be to combine the two original pages into one. This represents that new combined element which was a substantial piece of work on its own as well as the few other small changes suggested by him and the uproxy team, including:

Turning the user and server details dialogs into modals (giving them black background).
Making the icon at the top right of the details dialogs into an X instead of the expand-close.
Letting users click anywhere on an item to open its details dialog (rather than just the expand-open icon). This also implies that clicking the enable/disable button does NOT open the dialog, just as it was previously.
Removing public/private keys from the user details dialog and adjusting the buttons there as appropriate.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/71)


blueandgold:
LGTM  Cool!  Please make sure the changed server form also looks good on the setup page.

xwsxethan:
The server form on the setup page was not changed at all, so no worries!

PR #73, opened 2016-03-16
Rename left icon container.

blueandgold:
Changing the name here to something that's nicer, and is more consistent with the "headerContainer".

When testing this, I got the error: “Uncaught ReferenceError: Polymer is not defined” thus unable to see the left icon.  This can be fixed by importing polymer.html.  Would you know why?  This is especially odd since none of the other custom elements need this.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/73)

### DIFF
--- a/ufo/static/leftIconContainer.html
+++ b/ufo/static/leftIconContainer.html
@@ -1,4 +1,6 @@
-<dom-module id="form-container-with-left-icon">
+<link rel="import" href="bower_components/polymer/polymer.html">
+
+<dom-module id="left-icon-container">
   <style is="custom-style">
     #container {
       display: table;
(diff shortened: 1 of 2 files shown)


xwsxethan:
Hmmm, I guess without any imports, it doesn't know what elements to vulcanize into here, so you wouldn't have anything that says to get Polymer. If you added any other element, it would recursively depend on Polymer, thus picking it up for you. Still a bit odd, since in other cases it seems like vulcanization is able to work it out. In any case, the change looks fine to me. :+1: 

---

Reviewed 3 of 3 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/73#-:-KD3xCe-tDKeoNGVMJXb:-1500820018)*

PR #75, opened 2016-03-17
Updating Chrome Policy Format for Proxy Servers

xwsxethan:
Updating Chrome Policy to match uProxy's naming convention for settings as well as include proxy servers as dictionary objects with ip address and public key as properties (rather than just the public key).

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/75)


blueandgold:
Cool!  :+1:

PR #76, opened 2016-03-17
Fix users being adding multiple times.

blueandgold:
Initially I tried to to take care of this on the application level, i.e. using an extra if conditional to check for duplication.  Then, I played around with doing this on the database level, which seems to be much cleaner & robust.

The only trade-off here is that we need to save the user to the database one at a time, rather than by batch.  Because if one user is being added more than once, then whole batch will fail.  Can you see anywhere else where this might matter?

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/76)

### DIFF
--- a/ufo/database/models.py
+++ b/ufo/database/models.py
@@ -3,6 +3,7 @@ import StringIO
 from Crypto.PublicKey import RSA
 from paramiko import hostkeys
 from paramiko import pkey
+import sqlalchemy
 
 import ufo
 from ufo.services import ssh_client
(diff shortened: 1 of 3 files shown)


xwsxethan:
This looks good to me. Should we do the same thing for Proxy Servers (with uniqueness based on IP address perhaps)? :+1: 

---

Reviewed 3 of 3 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/76#-:-KD8pLCkY9KD1C8fnthE:1970194361)*

blueandgold:
Yes, indeed, good point about doing the same for the proxy servers.  Will follow-up with that separately.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/76#-:-KD9VVlO7bkkuo6KWnfe:-1473876407)*

PR #79, opened 2016-03-21
Making the Dropdown UI Work

xwsxethan:
This PR got slightly out of hand since a lot of cleanup work needed to be done to bring all these pieces together, so my apologies for the monolithic review. It basically breaks into a few major pieces:

First, the resource dictionaries were moved out of their respective handlers and into one central service. Each dictionary was then added into jinja environment globals so that they are available on every page without having to be manually added to each handler. I refactored each handler as necessary to pull these out. I also added a generic "landing" resource dictionary for resources used across the landing page, such as the dropdown menu and search.

Second, I created an admin module with basic list, add, and delete methods for interacting with admin users. I also added tests for each of these. I still need to add tests for the actual database methods, but this allows us to interact with admins outside of the setup flow.

Third, I wrote the actual dropdown UI element and wired it up to all the necessary pieces. Adding an admin is bundled as a dialog along with the dropdown UI dialog. It shows a status message for whether adding the admin was actually successful or not. Clicking the settings button redirects to the settings component on the setup page. We can move that component around if need be and even add it to the landing page if necessary, possibly making it pop up when requested. I'm open to suggestions on this, but for now the link works well. Finally, logging out posts to the logout handle and redirects the user back to the login page.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/79)


blueandgold:
Really nice overall!  Just a couple of questions on performance & structure related to the resource_dicts.

---

Review status: 0 of 23 files reviewed at latest revision, 5 unresolved discussions.

---

*[ufo/__init__.py, line 81 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQEgZuHQ7Yjdg11QeM:-KDQEgZvIxhtWNm71JF3:-1193355836) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/__init__.py#L81)):*
Hmmm, I am wondering how this might impact performance for every request:

(1) for non-UI, ajax calls, which might not need the resource_dicts, would all these resources_dicts be generated?

(2) for UIs, would all these resources_dicts be necessary?  for example, the landing page doesn't need get_oauth(), get_settings(), etc?  likewise for the setting page.

---

*[ufo/services/resource_dicts.py, line 1 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQJ6Ovj6K2MoAIMklN:-KDQJ6Ovj6K2MoAIMklO:-1907333309) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/services/resource_dicts.py#L1)):*
This resource_dicts.py feels kind of bloated and will seem to only grow more so in the future.  How do you feel about breaking this module down by creating a "services/resources" package, and then create individual user_resources.py, proxy_server_resources.py, etc?  So then inside, there would just be one simple get() method?

I would also not have the return type in the module naming, i.e. not use "_dict".

This means that if we really want set_jinja_globals() (but see my comment above on performance first), then you can create a resources_provider.py module and then stuff set_jinja_globals() in there.

---

*[ufo/services/resource_dicts.py, line 141 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQKhwRS_ya46eRsJS5:-KDQKhwRS_ya46eRsJS6:-644139264) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/services/resource_dicts.py#L141)):*
nit: seem like not using return_type in the naming would be nicer.

---

*[ufo/services/resource_dicts.py, line 185 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQGV-RUfIkYxZxVjzK:-KDQGV-RUfIkYxZxVjzL:-1834508603) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/services/resource_dicts.py#L185)):*
Should we add some unit tests for all these get methods to make sure the dicts being returned are sane?  Or is it sufficient to just rely on testing these upstream?  i.e. whatever are using these resources are configured / behaves properly?

---

*[ufo/services/resource_dicts.py, line 185 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQIRFgPlEDEQcwhRw6:-KDQIRFgPlEDEQcwhRw7:981709014) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/services/resource_dicts.py#L185)):*
nit: Can "with_add_flow" be called as "has_add_flow"?  So that it's more in sync with the actual key name?

```python
if not has_add_flow:
  proxy_server_resource_dict['hasAddFlow'] = False
```

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-:-KDQL5WPKxNNlbeUz3dZ:-739128269)*

xwsxethan:
Reviewed 23 of 23 files at r1.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[ufo/__init__.py, line 81 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQEgZuHQ7Yjdg11QeM:-KDQOat19wNAeWOxY_6I:-910559258) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/__init__.py#L81)):*
So for starters, I haven't noticed any impact on performance. Load time is still well under 1 second and hasn't seemed to be any different than usual. I think this is all handled server side which should intelligently not insert anything extra which isn't on the page once the template is rendered. I may be able to switch this to before_first_request so it literally only gets called on the very first request (since you don't need to set the globals multiple times anyway), though I haven't tried that yet. As for your direct questions:

(1) The AJAX calls use flask.response rather than render template, so jinja should not be touched during AJAX, thus no rendering of these resource dictionaries. They should only be rendered on requests that result in templates, which are just /setup and /.

(2) Not every resource dictionary will be necessary on every page, but as mentioned above, the server should handle this intelligently. So far, every page will need the generic landing one in order to handle the dropdown menu and search bar correctly.

The other part to think about here is that these are basically constants, analogous to a JSON file of language-specific text (as they will soon be used this way).

---

*[ufo/services/resource_dicts.py, line 1 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQJ6Ovj6K2MoAIMklN:-KDQQ3pj1LVta5bnuUCu:-1244263017) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/services/resource_dicts.py#L1)):*
I would prefer not to shard this out unless there will be any added functionality by separating them. In fact, I'd like to remove the with/has add flow portion and make that just a property that is set within the template, so these can be more straightforward instead of having separate with/without add flow versions.

I think of them as glorified constants in this use case. However, I cannot actually write them directly as constants, because calling flask.url_for won't work outside of the context of a request (so setting up the constant outside of a function in the module, means that the constant will get initialized whenever any file reads from __init__.py which reads from this causing a slew of problems). From that perspective, having a handful of files that each just provide one constant seems like overkill.

I could move these back to their respective handler files so we aren't creating new files just for the constants, and then just make this file be the provider that calls set_jinja_globals if that would work? I agree that removing _dict from the name seems fine however.

---

*[ufo/services/resource_dicts.py, line 141 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQKhwRS_ya46eRsJS5:-KDQSL3DGM9Nj2awdvie:-456589151) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/services/resource_dicts.py#L141)):*
Agreed, will fix tomorrow, just commenting so I know what to do tomorrow.

---

*[ufo/services/resource_dicts.py, line 185 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQGV-RUfIkYxZxVjzK:-KDQSC_ouz3kCzWnfFDx:709175985) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/services/resource_dicts.py#L185)):*
I would think these should be fine without unit tests since they are basically constants, but the UI should still be exercised from a functional test perspective, which would in turn ensure that these are constructed properly. The updates for WebDriver are another TODO of mine, so I will follow up with those shortly.

---

*[ufo/services/resource_dicts.py, line 185 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQIRFgPlEDEQcwhRw6:-KDQSxi1pvfL9dBiDJJa:-1321667712) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/services/resource_dicts.py#L185)):*
Per my comment above, I'd actually like to remove the with/has add flow portion entirely and make that just a property that is set within the template. So in setup.html for example, in addition to passing proxy_server_resources to the left-icon-container, it would also pass has-add-flow="false" so that logic is separate from the other resources, and only one version of the resource dictionary is necessary. I can still name it "has-add-flow" as you mentioned so it is consistent with a boolean name. Similarly for user resources as well. I'll follow up on this tomorrow when I come back to this PR.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79)*

blueandgold:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[ufo/__init__.py, line 81 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQEgZuHQ7Yjdg11QeM:-KDQ_6qBKRyUnXTuV8JO:-1587092693) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/__init__.py#L81)):*
It seems to me as if @before_request will cause set_jinja_before_request() to be triggered on every request to flask, even if no template rendering is involved.  (though the documentation is not explicitly clear, but that's what the naming implies)

For example, when downloading the chrome policy, the request goes to flask and will cause set_jinja_before_request() to be triggered, even if the no template rendering is involved.  And this is something that we should avoid.  There's gotta be some impact involved.  :)

https://github.com/uProxy/ufo-management-server-flask/blob/99c1907cd54e39d85c21d214263a8e22c45fae2d/ufo/handlers/chrome_policy.py#L75

@before_first_request seems better if these resources are all static.

Let's chat more about this if necessary tomorrow.  Maybe there's some crossed wire that we can clear up.

---

*[ufo/services/resource_dicts.py, line 1 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQJ6Ovj6K2MoAIMklN:-KDQcNb4vevwWpsyt83B:-1226342583) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/services/resource_dicts.py#L1)):*
Leaving them here is more preferred than in the handlers.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79)*

xwsxethan:
Reviewed 5 of 8 files at r2.
Review status: 23 of 25 files reviewed at latest revision, 5 unresolved discussions.

---

*[ufo/__init__.py, line 81 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQEgZuHQ7Yjdg11QeM:-KDTcKYmjix31IRa914u:927289275) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/__init__.py#L81)):*
So I verified that before_first_request works as expected and can be substituted for before_request in this case. Since they are now only being set once, the global variables shouldn't adversely impact performance, even if present in the environment on other requests, because they are only exercised when a template is rendered in the response.

---

*[ufo/services/resource_provider.py, line 185 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQIRFgPlEDEQcwhRw6:-KDTh6h_j1nAj2KNixgH:1186139525) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/services/resource_provider.py#L185)):*
Done. Note that this will look a little funny since boolean properties in Polymer are based on presence. For instance, passing like below:

has-add-flow="false"

will actually set hasAddFlow to true since has-add-flow is present. To set it to false, do not explicitly set anything. To set it to true, set has-add-flow to anything, but preference is for true just for clarity. I've updated the templates as necessary to reflect this.

---

*[ufo/services/resource_dicts.py, line 1 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQJ6Ovj6K2MoAIMklN:-KDTdqXpdJ7aZ-tUgyQe:413778086) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/services/resource_dicts.py#L1)):*
I switched them over to be private functions to more closely align with the use case I'm intending.

---

*[ufo/services/resource_dicts.py, line 141 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQKhwRS_ya46eRsJS5:-KDTdop6ausa_XmQpZGp:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/services/resource_dicts.py#L141)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79)*

blueandgold:
:lgtm: 

---

Review status: 23 of 25 files reviewed at latest revision, 4 unresolved discussions.

---

*[ufo/__init__.py, line 81 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQEgZuHQ7Yjdg11QeM:-KDUMTg_fQZIUlnNXhLo:746934339) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/__init__.py#L81)):*
Awesome!

---

*[ufo/__init__.py, line 84 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDUM6RHft3hAKU2tiO5:-KDUM6RHft3hAKU2tiO6:1753780109) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/6a8c77d9782cbda76cb3969636b5dbda8efbea7a/ufo/__init__.py#L84)):*
Awesome, this is very nice to read here!

---

*[ufo/services/resource_provider.py, line 185 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDQIRFgPlEDEQcwhRw6:-KDULl61ZGXt2slhan8M:-276167100) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/b484e707ba8f82f94af5a9e928fed75d6997f265/ufo/services/resource_provider.py#L185)):*
Thanks, noted.  It can be confusing.

---

*[ufo/services/resource_provider.py, line 275 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDUI4NrLIYo0P78FZVo:-KDUI4NsjZhLW7naD_iD:-1411649999) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/6a8c77d9782cbda76cb3969636b5dbda8efbea7a/ufo/services/resource_provider.py#L275)):*
nit: This is really nit-picking (and okay, to take a pass here) on something that's already in place, but looking at this again, it would help the cause of readability here if all these get methods include "resources" in the naming.  Otherwise, at first glance, it seems we are getting chrome policies and then setting it into resources.

```python
ufo.app.jinja_env.globals['policy_resources'] = json.dumps(_get_policy_resources())
```

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-:-KDUMVfxXQ0UCWY9FYmt:980077423)*

xwsxethan:
Reviewed 3 of 8 files at r2, 1 of 1 files at r3.

---

*[ufo/services/resource_provider.py, line 275 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79#-KDUI4NrLIYo0P78FZVo:-KDUQxppEzitLZDjc7LW:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/6a8c77d9782cbda76cb3969636b5dbda8efbea7a/ufo/services/resource_provider.py#L275)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/79)*

PR #80, opened 2016-03-22
Update Web Driver Functional Tests to Work with the New UI

xwsxethan:
This is at a decent enough state that I'm going to go ahead with a review, just to keep the PR from becoming monolithic when I write a bit more. The changes here are summarized below:
- Update web driver tests to use the new login/logout module so that existing tests will pass (no tests for reaching pages without logging in yet, but that will come later).
- Remove user page and sidebar tests so they won't run since those components no longer exist.
- Add tests for adding/deleting users/servers from the landing page for a base line of functionality. I haven't really touched the setup page tests yet, but that will also come later. I stubbed out a module to fill in.
- Add id tags to various elements to facilitate the web driver testing process so that specific elements can be easily found when necessary.
- Slight tweak to the login process: when a user is already logged in, navigating to /login manually (for whatever reason) redirects to the landing page. It doesn't make much sense to be able to login again while already logged in, so this helped me not put the server in a weird state if a user is already signed in.
- Update ui_test_suite.py and web_driver to use a username field instead of email since that seemed more appropriate for the current implementation. Once I get into the weeds with the setup page tests, I will also add a domain field so that we can test setting a domain and adding users that way. This will be difficult to get working since it fundamentally alters the state of the management server and will be hard to revert after the test, so I'm leaving that as a final item and avoiding for now.

To invoke the tests, make sure to re-vulcanize and push your files (either to a local 0.0.0.0 instance or a remote one), then start the tests with this command:

`./web_driver.sh test your_server_base_address_goes_here username_goes_here password_goes_here`

Example:

`./web_driver.sh test http://0.0.0.0:5000 ethanAdmin password_goes_here`

Note that the server base address does NOT have a trailing slash character and that the username and password passed in must already be setup along with the domain on the server pointed to.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/80)


tests/ui/base_test.py
```diff
@@ -30,3 +54,184 @@ def setContext(self):
...
+    # Find the user and navigate to their details page.
+    self.driver.get(self.args.server_url + flask.url_for('landing'))
+    landing_page = LandingPage(self.driver)
+    user_list_item = landing_page.GetElement(LandingPage.USER_LIST_ITEM)
+    user_listbox = user_list_item.find_element(*LandingPage.GENERIC_LISTBOX)
+    user_item = self.find_item_in_listing(
+        user_listbox, BaseTest.TEST_USER_AS_DICT['name'])
+
+    if user_item is None:
+      if raiseException:
```
blueandgold: nit: "shouldRaiseException"; otherwise, i was actually looking for what "raiseException" is.

tests/ui/base_test.py
```diff
@@ -30,3 +54,184 @@ def setContext(self):
...
+
+    # Click delete on that user.
+    details_modal = user_item.find_element(*LandingPage.DETAILS_MODAL)
+    WebDriverWait(self.driver, 10).until(
+        EC.visibility_of(details_modal))
+    delete_button = details_modal.find_element(*LandingPage.USER_DELETE_BUTTON)
+    delete_button.click()
+
+    # Wait for post to finish, can take a while.
+    WebDriverWait(self.driver, 20).until(
```
blueandgold: QQ: why is it that we setting a timeout of 20 here, but 10 in line 103?  Why not pick something consistent, say 30 seconds across the board?  I'm just worried that something short like 10 might cause flakiness in case the server is loaded or otherwise slow.  30 usually is what people have settled on, since it's long enough, yet the test will move on when the EC is fulfilled.

tests/ui/base_test.py
```diff
@@ -30,3 +54,184 @@ def setContext(self):
...
+    items = listing.find_elements(By.TAG_NAME, 'paper-icon-item')
+    for item in items:
+      # This can technically return multiple, but it will only return one.
+      div = item.find_elements(By.CLASS_NAME, 'first-div')[0]
+      strong = div.find_elements(By.TAG_NAME, 'strong')[0]
+      if name.lower() in strong.text.lower():
+        return item
+    return None
+
+  def add_test_server_helper(self, containingElement):
```
blueandgold: nit: why camel case here?

tests/ui/landing_page_test.py
```diff
@@ -1,31 +1,113 @@
...
+    landing_page = LandingPage(self.driver)
+    for element_by_id in LandingPage.BASE_PAGE_ELEMENTS:
+      base_page_element = landing_page.GetElement(element_by_id)
+      self.assertIsNotNone(base_page_element)
+
+    for element_by_id in LandingPage.LANDING_PAGE_ELEMENTS:
+      landing_page_element = landing_page.GetElement(element_by_id)
+      self.assertIsNotNone(landing_page_element)
+
+  def testManualUserAddFromLanding(self):
```
blueandgold: nit: maybe the wording can be "testManuallyAddUserFromLandingPage()"; do we still have the constraint on how long namings can be?

tests/ui/landing_page_test.py
```diff
@@ -1,31 +1,113 @@
...
+    self._add_test_user_from_landing_page()
+
+    self.assert_test_user_presence_on_landing_page(True)
 
-    self.driver.get(self.args.server_url)
+    self.remove_test_user()
 
+    self.assert_test_user_presence_on_landing_page(False)
+
+  def testServerAddFromLanding(self):
```
blueandgold: nit: if we're changing above, maybe this can be "testAddServerFromLandingPage()"

blueandgold:
:100:  Very nice!  The only thing I would like to see is that we have a login_page_test that to test incorrect logins (i.e. incorrect logins can not access the app).

tests/ui/base_test.py
xwsxethan: Done.

tests/ui/base_test.py
xwsxethan: I had just left the origin ones at 10 since that's what they had always been. 20 was necessary since the add user post does tend to take a while at random times. A default time works better though, and setting it to something sensible like 30 sounds good. Done.

tests/ui/base_test.py
xwsxethan: I'm honestly not sure which we should be following anymore. We were doing lower_with_underscore for code and camelCase for tests at one point, so I can do that if it is preferred. Here the lower_with_underscore named methods were for helpers and other random methods that weren't self-contained tests or the setUp/tearDown methods which have to be specially named. I will gladly change to whatever is preferred.

tests/ui/landing_page_test.py
xwsxethan: We should have the limit yes, but the health checks haven't been working for a while, and we have long names elsewhere. I don't think long names should really be that much of a problem if they are necessary. Done.

tests/ui/landing_page_test.py
xwsxethan: Done.

tests/ui/base_test.py
blueandgold: Yes, the casing is confusing.  Here specifically, "containing_element" would be better & consistent.

xwsxethan: Done and I updated the methods to all be camelCase for consistency.

xwsxethan:
As per your comment on login page tests, yes, I agree. Those will come later. I was more worried about testing the core pieces for now and forking everything else into separate PR's so that this didn't get too obnoxious in size. Thanks for the review though, as always!

PR #81, opened 2016-03-23
Implement new error notification UI.

blueandgold:
Please focus on the mechanics.  There will be some more clean-ups such as style the actual error notification UI and adding more error listeners in a follow-up PR.
- added a new polymer element for error notification
- this new element has a listener for error message events
- this new element is added to the base.html
- reworked the backend so that it return the error messages as json
- added listener for the json response, which will dispatch the error message event to the error notification element (is this step really necessary?  do you see a way for the base element to pick up the form response?)

screenshot from 2016-03-23 16 21 27

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/81)


xwsxethan:
Nice work with the backend changes and the dialog. Looks good to me so far! As you mentioned, we would want more listeners such as in the user, oauth, chrome policy, settings, and search components in the UI as well as more wrapped responses on the backend, but I assume those will follow once this is proven out.

My only major comment is how we're differentiating an error response from a success response. Does the server actually throw a 500 status code that you can see in dev console, or is that code just a property of the JSON object while 200 success code is returned? I can't tell if the flask.abort code gets overwritten by the error handler with flask.jsonify or not. 

If the actual error code is present, then you should be able to use the iron-form-error listener instead of iron-form-response (see here: https://elements.polymer-project.org/elements/iron-form#event-iron-form-error) to accurately differentiate and have different handlers for each case. Otherwise, we'll need to open up the response in the existing handlers, check if there is an error code present, and trigger your error notification if so, or continue with the default handler if not. It would be a lot cleaner and easier if the status code gets preserved, but I'm not sure if that's exactly possible with flask's wrapping.

---

Reviewed 7 of 7 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[ufo/services/custom_exceptions.py, line 25 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/81#-KDdIqiVn9F-1pqz6P0a:-KDdIqiVn9F-1pqz6P0b:-1889252149) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2221df3618f16192bfea80ba33a35d441d4ba8c5/ufo/services/custom_exceptions.py#L25)):*
We should eventually move these messages out to the resource dictionaries so that they can be i18n-ed, but I'm fine with just leaving a TODO here for followup.

---

*[ufo/static/errorNotification.html, line 29 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/81#-KDdJvCNV0xDnbGKwTt7:-KDdJvCNV0xDnbGKwTt8:1724094774) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2221df3618f16192bfea80ba33a35d441d4ba8c5/ufo/static/errorNotification.html#L29)):*
If there's only one error notification per page, then this is fine, but you could also use the following (just an fyi, no need to change it):

`this.$.errorNotificationDialog.open()` or similar with - instead of camel case (can never remember which it is with Polymer).

---

*[ufo/static/serverAddForm.html, line 72 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/81#-KDdJb4xDbUx8qbU1O25:-KDdJb4xDbUx8qbU1O26:-1053061379) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2221df3618f16192bfea80ba33a35d441d4ba8c5/ufo/static/serverAddForm.html#L72)):*
The parsePostResponse function below is already a form response listener, so you should be able to merge these two and have an if else block to differentiate success from failure based on the presence of our custom error code if we're receiving 200's still from the backend. If not (and we're actually getting 4XX or 5XX codes from the server), then you can actually modify the listener above to use iron-form-error which would be a more appropriate handler. See here: https://elements.polymer-project.org/elements/iron-form#event-iron-form-error

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/81#-:-KDdKIw7pq4hc1sXnVxG:-1540287448)*

xwsxethan:
Oh and just noticed this, you'll also want to fix whatever tests are breaking obviously.

---

Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/81#-:-KDdN-tEGviXJQg-ICQL:1147224990)*

blueandgold:
Thanks for the review!  Yup, iron-form-error listener was the first thing that I looked at.  But because the backend is intercepting the errors and wrapping the error response, the returned status will be a 200, thus the iron-form-error listener won't fire.  I actually found a way to overcome this by forcing the error_handler to return the actual error code.  This looks cleaner to me, so let me know how you think.

---

Review status: 2 of 10 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[ufo/services/custom_exceptions.py, line 25 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/81#-KDdIqiVn9F-1pqz6P0a:-KDdWrXj1yXpDP21rngB:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2221df3618f16192bfea80ba33a35d441d4ba8c5/ufo/services/custom_exceptions.py#L25)):*
Done.

---

*[ufo/static/errorNotification.html, line 29 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/81#-KDdJvCNV0xDnbGKwTt7:-KDdZ06lg89ELd-B0_XL:187265598) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2221df3618f16192bfea80ba33a35d441d4ba8c5/ufo/static/errorNotification.html#L29)):*
Changed to what you suggested, it's cleaner.  Yah, the casing is starting to get confusing.  :(

---

*[ufo/static/serverAddForm.html, line 72 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/81#-KDdJb4xDbUx8qbU1O25:-KDdbi1E-e25OasMhaZq:449120002) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2221df3618f16192bfea80ba33a35d441d4ba8c5/ufo/static/serverAddForm.html#L72)):*
Ah, okay.  I had missed this in the code.  But as mentioned above, changing this to the error listener.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/81#-:-KDdccdw6MLMbJKckQlY:1579207186)*

xwsxethan:
This looks much better now! I assume the changes to other front end handlers will be a separate PR, but looks good to me. :lgtm: 

I think you need to resync your repo though since merge is blocked. Just checkout master, pull, checkout your branch, and merge is probably easiest.

---

Reviewed 8 of 8 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[ufo/handlers/proxy_server.py, line 50 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/81#-KDeIpxTKkBcc1h1musK:-KDeIpxTKkBcc1h1musL:-1828097643) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/7fc1389fd0cb030ab8c8fd1459746a87a26a6240/ufo/handlers/proxy_server.py#L50)):*
I think you now want to switch this format back to match the others, or else the admin and user error formats are wrong.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/81#-:-KDeIz-1e3j3n8whSr5z:956353681)*

blueandgold:
Correct, will apply the error handling to the other frontend pieces after this PR.  I might try to style to actual error notification overlay first, though.

Yes, merging.

---

Review status: 8 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[ufo/handlers/proxy_server.py, line 50 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/81#-KDeIpxTKkBcc1h1musK:-KDeLQ621Rx9mJ1H_90Y:-1378396785) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/7fc1389fd0cb030ab8c8fd1459746a87a26a6240/ufo/handlers/proxy_server.py#L50)):*
Good catch.  The others need to match this here.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/81#-:-KDeMmSTagKY1ZEaJLEn:-721408094)*

PR #83, opened 2016-03-24
Make the new error notification UI look & behave like the mock.

blueandgold:
Please see the attached screenshot, which should look & behave like the mock.

The one detail that I spent a lot of time is to make sure the window doesn't close before the CLOSE button's ripple effect has finished.  This is accomplished by specifying the on-transitionend attribute.  Otherwise, it feels the window closes too quickly before the button ripple has finished.

screenshot from 2016-03-24 16 38 48

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/83)


xwsxethan:
:+1:

PR #85, opened 2016-03-28
Hook in the new error notification for users and admins.

blueandgold:
- This PR hooks in the new error notification when trying to add users or admins more than once.
- I attempted to hook this in for other exceptions (SetupNeeded, NotLoggedIn), but I can't find an easy way register an error listener/handler on the frontend for these (like I can for a form-error or ajax-error).  So, they will continue to route to the error page.  (Is there any other ideas?  This is one thing I think where angular is superior to polymer.)
- Will follow-up to add webdriver tests.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/85)

### DIFF
--- a/ufo/handlers/user.py
+++ b/ufo/handlers/user.py
@@ -179,7 +179,6 @@ def add_user():
     except custom_exceptions.UnableToSaveToDB as e:
       flask.abort(e.code, e.message)
 
-
   return user_list()
 
 @ufo.app.route('/user/delete', methods=['POST'])
--- a/ufo/static/ufoDropdownMenu.html
+++ b/ufo/static/ufoDropdownMenu.html
@@ -191,6 +191,9 @@
           notify: true,
         },
       },
+      listeners: {
+        'iron-form-error': 'handleFormError',
+      },
       getXsrfToken: function() {
         return document.getElementById('globalXsrf').value;
       },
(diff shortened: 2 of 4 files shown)


xwsxethan:
This PR looks fine, please go ahead whenever with submitting. I wanted to ask though: what do you mean with the other exceptions, SetupNeeded and NotLoggedIn? When are they actually thrown vs just handled as redirects to the setup or login module? If they are redirects, then that seems fine. If we're actually getting an error when trying to do something, as in post to a form, then we're probably in a very weird state which shouldn't happen.

---

Reviewed 4 of 4 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/85#-:-KDzjomyWcXqE4Ujxuuc:-1034186165)*

blueandgold:
Well, what I meant was that I'm not able to display the exceptions like SetupNeeded in the new error notification UI.  Using the same SetupNeeded example, it is handled and raised correctly by the backend.  But nothing on the frontend is picking up what the backend is sending.  Maybe we can talk about it more offline, and think of something.  Will merge this.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/85#-:-KE2BbFCiDFUGwBT7YwI:-1075143972)*

PR #87, opened 2016-03-29
Webdriver test for the new error notification UI.

blueandgold:
- This is a functional test to make sure the new error notification UI is properly wired up, can be opened, and then can be closed.
- Please note how the error_notification page object is structured to be self-contained, i.e. the locators and methods that acts on them are all together, rather than having them in base_test.py and layout.py.
- For another example, I encountered `self.addTestUserFromLandingPage()`
  But page object model would have expected something like `landing_page.addTestUser()`
  And also `setup_page.addTestServer()` instead.
- Hope I have used the correct common methods.  But if I missed anything, please let me know.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/87)

### DIFF
--- a/tests/ui/error_notification.py
+++ b/tests/ui/error_notification.py
@@ -28,7 +28,7 @@ class ErrorNotification(BaseDriver):
     except NoSuchElementException:
       return False
     return True
-  
+
   def is_displayed(self):
     """Whether the error notification is displayed (i.e. visible) on the page.
 
--- a/tests/ui/error_notification.py
+++ b/tests/ui/error_notification.py
@@ -1,0 +1,64 @@
+"""Module representing the error notification page object & functionalities."""
+
+from selenium.webdriver.common.by import By
+from selenium.common.exceptions import NoSuchElementException
+from selenium.webdriver.support import expected_conditions as EC
+from selenium.webdriver.support.ui import WebDriverWait
+
+from base_driver import BaseDriver
+from login_page import LoginPage
+
+
+class ErrorNotification(BaseDriver):
+  """Error notification page object and functionalities."""
+
+  ERROR_NOTIFICATION_ELEMENT = (By.ID, 'error-notification')
+  ERROR_NOTIFICATION_DIALOG = (By.ID, 'errorNotificationDialog')
+  ERROR_NOTIFICATION_CLOSE_BUTTON = (By.ID, 'errorNotificationCloseButton')
+  ERROR_MESSAGE_ELEMENT = (By.ID, 'errorMessage')
+
+
+  def is_present(self):
+    """Whether the error notification element is present on the page.
+
+    Returns: boolean, true if present, else false
+    """
+    try:
+      self.GetElement(self.ERROR_NOTIFICATION_ELEMENT)
+    except NoSuchElementException:
+      return False
+    return True
+  
+  def is_displayed(self):
+    """Whether the error notification is displayed (i.e. visible) on the page.
+
+    Returns: boolean, true if present, else false
+    """
+    error_notification_dialog = self.GetElement(self.ERROR_NOTIFICATION_DIALOG)
+    if error_notification_dialog.is_displayed():
+      return True
+    return False
+
+  def has_error_message(self, message):
+    """Whether the specified message is present in the error notification.
+
+    Returns: boolean, true if present, else false
+    """
+    is_present = EC.text_to_be_present_in_element(self.ERROR_MESSAGE_ELEMENT,
+                                                  message)
+    if is_present is False:
+      return False
+    return True
+
+  def close(self):
+    """Close the error notification, so that it's not displayed anymore.
+
+    Not catching the NoSuchElementException on purpose, so that the
+    exception will cause the test to fail.
+    """
+    self.GetElement(self.ERROR_NOTIFICATION_CLOSE_BUTTON).click()
+    # Need to wait for the button's ripple effect to complete.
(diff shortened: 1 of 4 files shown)


xwsxethan:
Looks good overall to me. :+1:  The only thing I would suggest is changing what you assert on for the specific message. I've tried to decouple all the other tests from assertions on specific strings so that we can avoid any increased maintenance cost if/when i18n rolls around.

However, as I'm typing this I realize that the error messages also aren't specified in a resource dictionary along with all the other text. Since it will end up going pretty deep down the rabbit hole to fix this and in turn find the i18n text on the page, just leave a TODO for now, and we can circle back to it after cleaning up the page model stuff.

As for the page model changes, I can take a look at refactoring some of these. I think your instincts here are correct and moving some functionalities from the base_test out to base_driver or layout will actually solve a lot of this. Since the functionality in base_test is shared between pages, to me that seems to make sense to be in layout, which is in essence the base page that every other page inherits from. The methods specific to landing/setup page can be moved to their actual pages and call into layout as necessary for the shared common methods. Renaming to remove "FromXPage" seems appropriate once those are done. I'll start a PR for this shortly and send it over to you.

---

Reviewed 3 of 3 files at r1.

---

*[tests/ui/error_notification.py, line 31 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/87#-KE3ZXpWImOz_uh--U8s:-KE3ZXpWImOz_uh--U8t:1099296581) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/f2f210902743984c446f8e580695a882cb47987e/tests/ui/error_notification.py#L31)):*
Nit: spacing.

---

*[tests/ui/error_notification.py, line 62 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/87#-KE3ZTjf_LeqBZaa5Eie:-KE3ZTjf_LeqBZaa5Eif:1848960251) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/f2f210902743984c446f8e580695a882cb47987e/tests/ui/error_notification.py#L62)):*
Agreed here.

---

*[tests/ui/error_notification_test.py, line 49 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/87#-KE3YieG02hiDl8ajB6D:-KE3YieG02hiDl8ajB6E:1099296581) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/f2f210902743984c446f8e580695a882cb47987e/tests/ui/error_notification_test.py#L49)):*
Nit: spacing.

---

*[tests/ui/error_notification_test.py, line 59 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/87#-KE3Z0iyUlR297wr9q7k:-KE3Z0iyUlR297wr9q7l:1026011264) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/f2f210902743984c446f8e580695a882cb47987e/tests/ui/error_notification_test.py#L59)):*
Hmmm, this might be a pain to change later for i18n. Should we consider another approach for determining what specific exception is raised? Like having error codes (numbers) that are language agnostic or being able to find the i18n-ed message passed into the page than make sure that message matches what's in the error notification?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/87#-:-KE3ZmtJXEPX9Mo1dduB:362787145)*

blueandgold:
Thanks for the review!  Please treat the webdriver clean-up as low-priority.  We can update them as we go along, and I'll help as well.  The most important thing is that we got lots of great tests in place that we won't have to do QA!  :)

---

Review status: 1 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/ui/error_notification.py, line 31 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/87#-KE3ZXpWImOz_uh--U8s:-KE3dHJlHjhtNlJIFByi:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/f2f210902743984c446f8e580695a882cb47987e/tests/ui/error_notification.py#L31)):*
Done.

---

*[tests/ui/error_notification_test.py, line 49 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/87#-KE3YieG02hiDl8ajB6D:-KE3dl2FkQiCLWZqxhgh:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/f2f210902743984c446f8e580695a882cb47987e/tests/ui/error_notification_test.py#L49)):*
Done.

---

*[tests/ui/error_notification_test.py, line 59 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/87#-KE3Z0iyUlR297wr9q7k:-KE3eB2zl5M1BakaDAuP:-354353501) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/f2f210902743984c446f8e580695a882cb47987e/tests/ui/error_notification_test.py#L59)):*
Good point here on i18n.  Added the TODO to make this testable for i18n.  I don't really have good answer here, as this depends how we go about implementing i18n, and how the correct localized strings are sent.  So, let's punt on this for now.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/87#-:-KE3em1CJ595X4aSnzp5:1224784194)*

PR #88, opened 2016-03-29
Functional Tests to the Settings Component

xwsxethan:
This new set of functional tests verifies that the links to the settings component are in fact valid and that changing a settings and saving it will in fact be respected. The test necessitated a means to easily change a given setting and reset it back. Those pieces will likely prove very useful in the future as more settings are added. This new test should follow the page object model as previously fixed in the other tests.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/88)


blueandgold:
Looks good overall.  Just some nits & questions. :lgtm: 

---

Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/ui/layout.py, line 53 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/88#-KEHz0R8PmiXOghzWJzi:-KEHz0R8PmiXOghzWJzj:-850270096) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2011296ca31e037eb10550441c1347987a23a226/tests/ui/layout.py#L53)):*
nit: Should we move locators like these into the page object?  It should still be re-usable from there.

---

*[tests/ui/settings_component_test.py, line 60 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/88#-KEI11NWWT0R8jQa_7i4:-KEI11NXaiFLpvyqYKsB:1862925864) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2011296ca31e037eb10550441c1347987a23a226/tests/ui/settings_component_test.py#L60)):*
nit: We should add a comment here to say that we are reloading the page so that backend changes are reflected in the frontend.

---

*[tests/ui/settings_component_test.py, line 65 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/88#-KEI1LxjqTSjChbCD9iG:-KEI1LxjqTSjChbCD9iH:-1004228571) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2011296ca31e037eb10550441c1347987a23a226/tests/ui/settings_component_test.py#L65)):*
Should or will other settings be included for testing as well?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/88#-:-KEI2B7A5PIcX9Rlkgy3:925309706)*

xwsxethan:
Reviewed 8 of 8 files at r1, 5 of 5 files at r2.

---

*[tests/ui/layout.py, line 53 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/88#-KEHz0R8PmiXOghzWJzi:-KEI81KmHrKUSYGfBUxA:755545747) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2011296ca31e037eb10550441c1347987a23a226/tests/ui/layout.py#L53)):*
For the settings ones specifically, yes. However, since we reuse the ID for the proxy server display template, user display template, and chrome policy display template, those don't make as much sense. I was following the same approach here in the event that we end up displaying the settings on the landing page at some point in the future. However, it isn't now so I've went ahead with moving it. Done.

---

*[tests/ui/settings_component_test.py, line 60 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/88#-KEI11NWWT0R8jQa_7i4:-KEI8xA-xkKDdFQoQ5N5:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2011296ca31e037eb10550441c1347987a23a226/tests/ui/settings_component_test.py#L60)):*
Done.

---

*[tests/ui/settings_component_test.py, line 65 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/88#-KEI1LxjqTSjChbCD9iG:-KEI8zMINC3k43pM15EY:1969229276) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/2011296ca31e037eb10550441c1347987a23a226/tests/ui/settings_component_test.py#L65)):*
Yes, but I held off on writing them for now. I tried to make it very easy to add those in later once we have more settings. Obviously, it would be fairly trivial to test the enforce network jail settings using this same code, but that didn't seem all that useful.

I left a todo in the other file for adding ways to change future settings (whatever they may be) since they likely won't be just toggle switches, but potentially something more complex, like an input field that is more configurable. I'll add the necessary test for whatever those settings may be once we implement them.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/88)*

PR #89, opened 2016-03-31
Refactoring Web Driver Functional Tests to Use the Page Model

xwsxethan:
This doesn't change any existing test at all. The only thing this (rather large, sorry) PR does is move functions around and do a little renaming to make things more straightforward. All helper code that manipulates the UI on a given page was moved to that page and out of its test. The "FromLandingPage" (or similar text for other pages) was dropped from calls such as `addUserFromLandingPage` since it will now be called like `landing_page.addUser`.

Additionally, I removed any unused imports as well as finally deleted the sidebar page and sidebar test since those no longer make sense with the new UI.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/89)


blueandgold:
:lgtm_strong:  This really makes the tests much easier to read, i.e. it's easier to tell what object we're exercising.  More significantly, this organization will make sense to external contributors and help them to have as easy a time as possible to write or update the webdriver tests.

One comment for consideration only.  It's okay to pass on it as the code is small and we have made a lot of progress here.

Again, my apologies for not catching this in the earlier reviews.  Sometimes, things like this don't become obvious until we have to work with it directly.

---

Review status: 0 of 17 files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/ui/layout.py, line 127 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/89#-KEDs0Zm8eJ8v_WUX0ag:-KEDs0Zm8eJ8v_WUX0ah:-751675194) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/6c420cd92cfacd2104fb841dc45f1f7133b72094/tests/ui/layout.py#L127)):*
In looking at these 2 add methods (for user and server) and how they are used as common from the landing & setup pages, do you think that it might be indicating that we should turn them into page objects?  i.e. a page object for add_user_form, and a page object for add_server_form?  So that we can have something like this:

```python
add_user_form.addTestUserOnceFormIsDisplayed()
add_server_form.addTestServerOnceFormIsDisplayed()
```

This is very similar to how error_notification is setup as a page object.  So, that error_notification is usable from either landing page or setup page, rather than being here should we test that from multiple pages.

I think that "OnceFormIsDisplayed" can be dropped from the naming, since the state of being displayed is implied before we can add anything.

```python
add_user_form.addTestUser()
add_server_form.addTestServer()
```

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/89#-:-KEDuma6aETPd5GBS7nf:-843119458)*

xwsxethan:
Reviewed 17 of 17 files at r1, 6 of 6 files at r2.

---

*[tests/ui/layout.py, line 127 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/89#-KEDs0Zm8eJ8v_WUX0ag:-KEHTglysWVJDe9kWV7a:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/6c420cd92cfacd2104fb841dc45f1f7133b72094/tests/ui/layout.py#L127)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/89)*

PR #91, opened 2016-04-04
Fixing Hover Text on Paper Icon Buttons for Details Dialogs

xwsxethan:
As it turns out, there is in fact a hover text for these buttons. I had attempted to set it before using the title property, but at some point I updated the name of the variable in the resource provider and didn't change the corresponding variable in the Polymer components, thus causing undefined.

This PR fixes that issue as well as cleans up the text, so the expand button says something like "Expand User" or "Expand Server" while the close button just says "Close".

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/91)


blueandgold:
Sweet! :+1:

PR #93, opened 2016-04-04
Adding Max Height Attribute to Scrollable Div so that it Doesn't Expand Behind the Buttons

xwsxethan:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/93)


blueandgold:
:+1:

PR #94, opened 2016-04-04
Making the Close Dialog Function Work for the Remove Flow

xwsxethan:
Although the initial logic for checking whether the add flow was present was correct, the add flow and the remove flow are actually both always present in the html, so searching for them should always return them (thus passing the if check), so long as the html is rendered properly. It is simply whether or not they are displayed they we are curious about. However, in this case when we are closing either, it doesn't hurt to close both since only one will be open at a time, and since you can't navigate directly from one to the other without going back to the menu. I've tested that the below works as intended now without leaving either dialog open beyond its expected life.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/94)


blueandgold:
:+1:

PR #95, opened 2016-04-04
Making Dropdown Menu and Associated Dialogs Have the Dark Backdrop

xwsxethan:
This should be more consistent with other dialogs within the app. These still are not 'modals' in the pure sense where clicking outside of them would do nothing (clicking outside of these will actually close them), which is more consistent with the typical dropdown behavior as the mock originally intended.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/95)


blueandgold:
Awesome!  :+1:

PR #96, opened 2016-04-04
Fixing Bouncing Search Bar

xwsxethan:
I placed the search bar and spinner within a layout div so that enabling/disabling the spinner (making it appear/disappear) does not cause the bouncing up and down. The bouncing was because the spinner would be rendered underneath the bar (underneath in the sense of lower y position, not z position), and both were attempting to fit into one box that could grow vertically. By placing them in a horizontal box, the spinner appears/disappears on the right of the search bar as intended, so nothing bounces.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/96)


blueandgold:
Nice!  :+1:

PR #99, opened 2016-04-04
Redirect Users to HTTPS if They Visit HTTP

xwsxethan:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/99)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -27,6 +27,7 @@
     "paper-spinner": "PolymerElements/paper-spinner#~1.1.0",
     "paper-tabs": "PolymerElements/paper-tabs#~1.2.4",
     "paper-toggle-button": "PolymerElements/paper-toggle-button#~1.0.14",
-    "paper-toolbar": "PolymerElements/paper-toolbar#~1.1.2"
+    "paper-toolbar": "PolymerElements/paper-toolbar#~1.1.2",
+    "platinum-https-redirect": "PolymerElements/platinum-https-redirect#~1.0.1"
   }
 }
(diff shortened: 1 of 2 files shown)


blueandgold:
:+1:  Fantastic -- I didn't know there is a polymer element for this!

xwsxethan:
I didn't either. They list a couple of other options for doing this server-side, but this seemed so easy that I felt like I should just go ahead with adding it.

jpevarnek:
Sorry to jump in late, but can you make this optional (so that it does not break any server it is deployed on that does not support https).

As a side note: redirecting to https should really be done in the server configuration, not by the application (301 redirect for all requests)

xwsxethan:
Hmmm, I'm not sure the element supports turning this on and off directly. I could probably wire up something fancy with AJAX to see if we can do a get on the https prefixed setup page and condition on that, but it would be a bit hacky. Another way would be to make this a setting that can be turned on and off similarly to the network jail. I can also just remove this entirely instead if it will be problematic.

As for your side note, yes I agree. I haven't poked around with actually configuring that yet, but this seemed very trivial to add in the interim. From a cursory search, these links seemed like the best way to use flask to do this:

Update url_for to generate https links: http://stackoverflow.com/questions/14810795/flask-url-for-generating-http-url-instead-of-https
Redirect all http traffic to https: http://stackoverflow.com/questions/32237379/python-flask-redirect-to-https-from-http

To turn these off would require a server setting, but then if you turn this on, is there a path to ever turn it back off without using https (which may not be supported as you've mentioned)?

PR #101, opened 2016-04-05
Https Redirect on the Server Side

xwsxethan:
This isn't ready just yet, but I'm doing some of the plumbing now. Going to test on nightly to see if the migration works as well as flipping to https.

---
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/101)


jpevarnek:
Tiny comment: should really be happening at the level that's actually serving the HTTPS cert (e.g. nginx)

xwsxethan:
Sorry for delay. So when I actually turned this on, it managed to infinitely redirect causing the page not to load. I'm still going to play around with it to figure out why that's happening, but I've left this on the back burner for the moment.

Do you have any familiarity with nginx or how we could hook into that on Heroku, @jpevarnek ?

xwsxethan:
Since we're now going to utilize nginx on Heroku as specified in a separate PR (https://github.com/uProxy/ufo-management-server-flask/pull/175), I'm going to close this out.

PR #102, opened 2016-04-06
Creating Baseline of Postgres Database Columns for Future Migrations

xwsxethan:
This sets up the Flask-Migrate tool and checks in the basic stubs needed to run it. From here out, we can just create migrations and apply them.

I'm particularly interested in feedback about what should be checked in here. I did this largely based on the work here: https://realpython.com/blog/python/flask-by-example-part-2-postgres-sqlalchemy-and-alembic/

Based on those examples and working with Heroku, checking in the actual migrations seems like the right idea so that we don't all have individual copies. However, there are some extra files that come with the setup which may not be necessary. It will likely take some experimentation to figure out what is needed vs not needed.

Anyway, in order to create future migrations (after you've setup postgres locally), you'll just do the following:
1. Make a change to the models.py file to generate new schema.
2. Run `python manage.py db migrate` to generate a migration for the new schema.
   Once you've check in your code, this migration will be shared so that other developers and heroku can all pick it up.
3. Run `python manage.py db upgrade` to apply the migration to your local database.

Similarly, run `heroku run python manage.py db upgrade` once the code is checked in and pushed to heroku in order to apply the migration there. I'm working on a means of automating this, by for now the manual approach can be used.

This approach should work for simple migrations, such as adding a new column, renaming a column, or deleting a column (which in my estimate will likely be a majority of what we encounter). If we need to do a complex migration where data might be copied to several different columns or otherwise split in some way, we will have to modify the migration file created under migrations/versions. There are already stubs there to do the basic operations which we would expand. There will be notes with all of this information to follow shortly.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/102)


xwsxethan:
@jpevarnek Please take a look as well. This is just the basic setup without actually migrating anything yet.

blueandgold:
Super cool, thanks for taking care of this!  I am not sure how everything works here exactly.  So, I'm looking forward to the forthcoming documentation and trying this out.  :lgtm: 

---

Review status: 0 of 9 files reviewed at latest revision, 2 unresolved discussions.

---

*[manage.py, line 8 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/102#-KEgMC2cwMl7Mli-gC6q:-KEgMC2dd3q1HeKIUc0N:-1363552224) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/6c22bceb591509aa50b07ca89f529714da1fffb7/manage.py#L8)):*
Is this a file that you created?  i.e. can we rename it to something more clear, "db_migration_manager.py" or "manage_db_migration.py"?  Can we also have a module docstring to describe what this is for?

---

*[migrations/README, line 1 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/102#-KEgMYkKgrdwefUBzJpU:-KEgMYkKgrdwefUBzJpV:-1569104934) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/6c22bceb591509aa50b07ca89f529714da1fffb7/migrations/README#L1)):*
nit: remove line return

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/102#-:-KEgMrNKyULv_580MspP:669520203)*

xwsxethan:
Since the file manage.py was renamed to db_migration_manager.py, in the sample commands listed on the PR, replace manage with db_migration_manager.

---

Reviewed 7 of 8 files at r1, 2 of 2 files at r2, 3 of 3 files at r3.

---

*[migrations/README, line 1 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/102#-KEgMYkKgrdwefUBzJpU:-KEgTrJC2pUPrfDf6LOI:-1562302190) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/6c22bceb591509aa50b07ca89f529714da1fffb7/migrations/README#L1)):*
This file was auto generated as part of the migration init process, so I actually didn't even make it. I think you mean add line return here since there wasn't one, but regardless done.

---

*[manage.py, line 8 \[r2\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/102#-KEgMC2cwMl7Mli-gC6q:-KEgTAM0P4BQn3v4CCIY:2050732852) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/6c22bceb591509aa50b07ca89f529714da1fffb7/manage.py#L8)):*
This was created following the example from https://realpython.com/blog/python/flask-by-example-part-2-postgres-sqlalchemy-and-alembic/ under the Local Migration section, but it will work with any name. Done and done. FYI for the commands to run this later on, replace manage.py with the new name obviously, as in `python db_migration_manager.py db migrate` .

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/102#-:-KEgUDVgsag25cLv5ua-:726016813)*

PR #107, opened 2016-04-07
Making Cancel Button Reset the Form on the Setup and Landing Pages

xwsxethan:
Also cleaning up the isSpinnerOn variable to just be loading like the rest.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/107)


blueandgold:
:+1:  nit: Not critical, but wondering if we should call the button "Clear" instead of "Cancel".

xwsxethan:
That makes more sense on the setup page where the form is present outside of a dialog and persists after you hit cancel/clear, but that doesn't really make sense on the landing page where cancel/clear resets the form and actually cancels out the entire dialog. In the future it may make more sense to fork these out more and make the subcomponents smaller and easier to combine.

For instance, if we pulled the actual form element itself out (separate from the dialog) then it would be easy to reuse that across pages and specify different text instead of having the same overall dialog on both pages.

PR #109, opened 2016-04-07
Remove Auto Validation for Forms (Validate on Submit)

xwsxethan:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/109)


blueandgold:
Can you please add a note on why we are doing this?

xwsxethan:
It was a request from Michael that we only validate on submit rather than automatically as the user types. Auto validate can lead to some interesting scenarios, for instance as you type an email, you're required to have text after the dot in say my_email_address@foo.com. So as you type, once you hit my_email_address@foo. , it will show an error until the com part is added. From a personal standpoint, I don't really have a preference either way, and we are still validating the fields, just not until submit is triggered.

blueandgold:
Got it, was just wondering the motivation for this change.  :)  :+1:

PR #110, opened 2016-04-07
Fixing Invite Code Display When There is No Server Available

xwsxethan:
Adding notification for why an invite code is not available and making the spinner stop so it doesn't go on indefinitely. Also fixing the server side so it doesn't throw an error.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/110)

### DIFF
--- a/ufo/handlers/user.py
+++ b/ufo/handlers/user.py
@@ -234,8 +234,12 @@ def user_get_invite_code():
   user_id = json.loads(flask.request.args.get('user_id'))
   user = models.User.query.get_or_404(user_id)
   invite_code = _make_invite_code(user)
-  invite_url = INVITE_CODE_URL_PREFIX + invite_code
-  code_json = json.dumps(({'invite_code': invite_url}))
+  code_json = None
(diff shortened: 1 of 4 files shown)


blueandgold:
Hmmm, why aren't the error notification being used?  Is there any consideration to do this over that?

xwsxethan:
I considered it, but ultimately error notification isn't used here since this is an automatic process rather than the user specifically asking for an invite code. As in, as soon as you open user details, we fetch the current invite code for you if you don't already have one. Clicking to create a new invite code is a different matter and should still use the error notification as normal, but this is just for the initial flow of populating the page. Does that sound good to you?

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/110#-:-KEmY8G2fHuAYrenaSxJ:-372251763)*

blueandgold:
Sounds good.  :+1:

PR #112, opened 2016-04-08
Make List of Users to Check Display Correctly on Setup Page

xwsxethan:
I tried to just show/remove the paper-dialog-scrollable element, with its contents always displayed, but that didn't seem to work correctly so I have to copy the form element into the if and if ! sections. This worked out though because the setup page version will need some more styling to look a little cleaner. I'll work more on that in a separate PR. This gets the form back to working for the moment though.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/112)

### DIFF
--- a/ufo/static/userAddForm.html
+++ b/ufo/static/userAddForm.html
@@ -38,7 +38,7 @@
         </form>
       </template>
       <template is="dom-if" if="{{!showInputFields}}">
-        <template is="dom-if" if="{{displayWithDialog}}">
+        <template is="dom-if" if="{{shouldDisplayWithDialog}}">
           <paper-dialog-scrollable>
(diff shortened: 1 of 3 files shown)


ufo/static/userAddForm.html
```diff
@@ -38,7 +38,34 @@
         </form>
       </template>
       <template is="dom-if" if="{{!showInputFields}}">
-        <paper-dialog-scrollable>
+        <template is="dom-if" if="{{displayWithDialog}}">
```
blueandgold: nit: would be nice to have boolean naming
```
<template is="dom-if" if="{{shouldDisplayWithDialog}}">
```

blueandgold:
:+1:

PR #118, opened 2016-04-11
Fixing Order of DB Items Returned for Display in List Queries

xwsxethan:
This simple change just ensures that the result of listing entities as dictionaries will always be in order by ID so that results aren't swapped around when a new list is returned.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/118)

### DIFF
--- a/ufo/database/models.py
+++ b/ufo/database/models.py
@@ -33,6 +33,7 @@ class Model(ufo.db.Model):
   Most method implementations are taken from
   https://github.com/sloria/cookiecutter-flask"""
   __abstract__ = True
+  id = ufo.db.Column(ufo.db.Integer, primary_key=True)
 
   def update(self, commit=True, **kwargs):
     """Update the given entity by setting the given attributes and saving.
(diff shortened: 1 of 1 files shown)


blueandgold:
:lgtm: Will I need to use the db migration tool on the db in my instance?  If so, guess it's a good chance to try the tool out.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/118#-:-KF5NWcROx0BYUc3OKRW:1599857334)*

xwsxethan:
No, you won't have to migrate anything. The schema actually doesn't change at all since each entity already had an ID field, but I added the ID field to the base abstract class to formalize that relationship. The other change was just in how results are returned.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/118#-:-KF5ONX2XwxC6-3si_tl:885991952)*

PR #119, opened 2016-04-11
Remove Access Column in List Items and Update Formatting

xwsxethan:
This removes the "Access Enabled/Disabled" text on the user listing on the landing page. It also is removed from the details views and the format is changed slightly to add more room for the email address field.

This change was part of Michael's suggestions for an easier to use enable/disable flow. See screenshot attached for new look:
ufolistitemscolumns

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/119)


blueandgold:
:+1:

PR #123, opened 2016-04-13
Using New Logo and Adjusting Styling for Appropriate Size

xwsxethan:
ufonewlogo

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/123)


blueandgold:
:+1: Looking at the logo, it looks to me it could be a bit bigger and more prominent?  But okay as it is.

PR #124, opened 2016-04-15
Making Forms Handle Enter Key Press to Submit

xwsxethan:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/124)


blueandgold:
+1 I was looking forward to this!

xwsxethan:
I have no idea how many times I've hit enter on the login screen thinking it would work, only to be disappointed. But no more!

PR #128, opened 2016-04-19
Stop installing npm packages globally/asking for permissions

jpevarnek:
Switch to using the local version of an npm package instead of
overwriting system-wide state, and definitely do not do anything as root
as part of a repository-local build step

---
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/128)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ install:
  - pip install -r requirements.txt
  - python setup_database.py
  - npm install
- - ./vulcanize.sh travis
 # command to run tests
 script:
  - python -m unittest discover -p "*_test.py"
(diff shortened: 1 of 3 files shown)


xwsxethan:
:lgtm: You'll also want to update .travis.yml I think. In fact, you may be able to remove the vulcanize line entirely since this happens as a result of travis picking up package.json and calling it there.

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/128#-:-KFjDTwfmZsQBJnJWaFz:323897296)*

PR #133, opened 2016-04-25
Add an extra step to describe how to make the public key usable for accessing the proxy server.

blueandgold:
There is a really important step that we should include to make the private key from /etc/ssh/ssh_host_rsa_key to be usable for accessing the server via ssh.  And that is the corresponding public key needs to be added to the /root/.ssh/authorized_keys.  This is not done by default when the proxy server is provisioned.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/133)


xwsxethan:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/133#-:-KGDs_WCKMdOiyBceaAx:1417088721)*

PR #135, opened 2016-05-05
Fixing the Copy Invite Code Button

xwsxethan:
We have to use a different event listener format for it to trigger properly in Chrome. The copy command doesn't actually pick up the necessary data to be copied, so we intercept that event, check that it is the event we just triggered, and let it pass if not or set the necessary data if so. Kind of a pain, but it works now.

Also, I cleaned up how our copy functions by having one static hidden node in the dom to put the content onto then copy from (rather than dynamically generating and destroying a node each time). I fixed the height on the buttons in the details dialog so that they are uniform as well. Finally, I added icons to the copy and create new code buttons.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/135)

### DIFF
--- a/ufo/static/globalHelperFunctions.js
+++ b/ufo/static/globalHelperFunctions.js
@@ -13,3 +13,19 @@ var submitUsersManually = function() {
   input.value = JSON.stringify(userArray);
   formElem.submit();
 };
+
+
+(function() {
+  'use strict';
+  document.body.addEventListener('copy', copyInviteCodeToClipboard, true);
+
+  function copyInviteCodeToClipboard(e) {
+    if (e.target !== document.body) {
+      return;
+    }
+    e.preventDefault();
+
+    var input = document.getElementById('hiddenCopyInput');
+    e.clipboardData.setData('text/plain', input.value);
+  }
+})();
--- a/ufo/static/globalHelperFunctions.js
+++ b/ufo/static/globalHelperFunctions.js
@@ -14,7 +14,13 @@ var submitUsersManually = function() {
   formElem.submit();
 };
 
-
+/**
+ * I had to do this outside of a custom element because of the way that the
+ * browser handles permissions for using the document.execCommand and
+ * clipboardData API's. In essence, it doesn't like how Polymer registers and
+ * fire event handlers and won't let you touch clipboardData from there, thus
+ * the Javascript is outside of Polymer.
+ */
 (function() {
   'use strict';
   document.body.addEventListener('copy', copyInviteCodeToClipboard, true);
--- a/ufo/static/userDetailsDialog.html
+++ b/ufo/static/userDetailsDialog.html
@@ -38,6 +38,9 @@
     .spacer {
       width: 20px;
     }
+    .buttons paper-button {
+      height: 40px;
+    }
   </style>
   <!-- TODO(eholder): Figure out how to make hover work with focused. -->
   <template>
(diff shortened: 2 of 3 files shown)


xwsxethan:
ufofixedcopybutton

Screenshot showing the new look.

blueandgold:
Nice overall, just a couple of items that I have questions on.

> ufofixedcopybutton
> 
> 
> 
> Screenshot showing the new look.
> 
> </details>


---

Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[ufo/templates/base.html, line 35 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/135#-KHLNTuBnRJ1iugTm23M:-KHLNTuC6ZaGTadt_yvb:1250593002) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/83cadc663b30ac5ef1dc565fe22a91dea42b73c3/ufo/templates/base.html#L35)):*
> ```HTML
>   #hiddenCopyInput {
>     position: fixed;
>     z-index: -1;
> ```

nit: would some other visibility attribute work better/cleaner rather than a z-index?  Such as:
`<input type="hidden" />`

---

*[ufo/templates/base.html, line 76 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/135#-KHLO-hxK8Uagayzfafq:-KHLO-hxK8Uagayzfafr:-1224224707) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/83cadc663b30ac5ef1dc565fe22a91dea42b73c3/ufo/templates/base.html#L76)):*
> ```HTML
> 
>   <script src="{{ url_for('static', filename='globalHelperFunctions.js') }}"></script>
>   <input id="hiddenCopyInput" value=""/>
> ```

Since there are associated js & styling for this, should this be better as a custom element?
<hiddenCopyInput></hiddenCopyInput>

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/135#-:-KHLOUm05QoTqK6HX3A_:-125365861)*

xwsxethan:
Reviewed 3 of 3 files at r1, 1 of 1 files at r2.

---

*[ufo/templates/base.html, line 35 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/135#-KHLNTuBnRJ1iugTm23M:-KHLeVwFbvjRlrwjk1_F:-1837700340) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/83cadc663b30ac5ef1dc565fe22a91dea42b73c3/ufo/templates/base.html#L35)):*
> nit: would some other visibility attribute work better/cleaner rather than a z-index?  Such as:
> `<input type="hidden" />`
> </details>
Done. I was trying to create this dynamically in Javascript before, so adding the attribute was less efficient than styling, then when I moved to statically defining it, I hadn't bothered to switch it over. This is better though, nice catch!

---

*[ufo/templates/base.html, line 76 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/135#-KHLO-hxK8Uagayzfafq:-KHLemzx0_ISn9nyn3DX:-1871604130) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/83cadc663b30ac5ef1dc565fe22a91dea42b73c3/ufo/templates/base.html#L76)):*
> Since there are associated js & styling for this, should this be better as a custom element?
> <hiddenCopyInput></hiddenCopyInput></details>
Because of the other change, there won't be any CSS now. Anyway, I had to do this outside of a custom element because of the way that the browser handles permissions for using the document.execCommand and clipboardData API's. In essence, it doesn't like how Polymer registers and fire event handlers and won't let you touch clipboardData from there, thus the Javascript is outside of Polymer. This should be fairly small and straightforward however, so hopefully it isn't too bad.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/135)*

blueandgold:
:+1:  Also, doesn't have to be done in this CL,, but it would be nice to have a functional test for this.

> Nice overall, just a couple of items that I have questions on.</details>


---

---

*[ufo/templates/base.html, line 76 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/135#-KHLO-hxK8Uagayzfafq:-KHLjpGl01L-G-16c3g5:-475322624) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/83cadc663b30ac5ef1dc565fe22a91dea42b73c3/ufo/templates/base.html#L76)):*
> Because of the other change, there won't be any CSS now. Anyway, I had to do this outside of a custom element because of the way that the browser handles permissions for using the document.execCommand and clipboardData API's. In essence, it doesn't like how Polymer registers and fire event handlers and won't let you touch clipboardData from there, thus the Javascript is outside of Polymer. This should be fairly small and straightforward however, so hopefully it isn't too bad.</details>
Okay, I think this deserves a comment in the code (maybe next to the JS function?) so that we can document / explain why we are doing this.  Your comment above is perfect for copy/paste as a code comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/135#-:-KHLkNN-_KjdnJBu-miD:1043796345)*

xwsxethan:
I'm not entirely sure that I can unfortunately. I looked at doing this actually, but it seems that web driver doesn't have access to the clipboard since that is OS level functionality. Basically, there's no hooks to see what just got copied to your clipboard. I may be able to hack around by sending ctrl + v while focused on an input box on the page then check that the input box has the value we wanted to be copied, but that seemed a bit backwards and generally not a good idea since there are a variety of ways it could fail.

> :+1:  Also, doesn't have to be done in this CL,, but it would be nice to have a functional test for this.</details>


---

Reviewed 1 of 1 files at r3.

---

*[ufo/templates/base.html, line 76 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/135#-KHLO-hxK8Uagayzfafq:-KHLl_8Sua8HNkgDvAln:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/83cadc663b30ac5ef1dc565fe22a91dea42b73c3/ufo/templates/base.html#L76)):*
> Okay, I think this deserves a comment in the code (maybe next to the JS function?) so that we can document / explain why we are doing this.  Your comment above is perfect for copy/paste as a code comment.</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/135#-:-KHLl5A2SeTJ3GfvUFdH:-382778774)*

blueandgold:
How about this for sending the keys (paste, specifically)?
http://stackoverflow.com/a/11755800

So, can we do a simple test like:

1: click the copy button
2: navigate to some field, maybe the search box
3: paste into the search box
4: assert what's in the search box is the copied invite code

This way, we don't have to check the clipboard.

PR #138, opened 2016-05-06
Create New Test for Editing a Server and Make Editing a Server Actually Work

xwsxethan:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/138)


tests/ui/landing_page.py
```diff
@@ -143,7 +143,43 @@ def removeTestServer(self, name, server_url, should_raise_exception=True):
...
+    # Click edit on that server.
+    details_modal = server_item.find_element(*LandingPage.DETAILS_MODAL)
+    WebDriverWait(self.driver, UfOPageLayout.DEFAULT_TIMEOUT).until(
+        EC.visibility_of(details_modal))
+    edit_button = details_modal.find_element(*LandingPage.SERVER_EDIT_BUTTON)
+    edit_button.click()
+
+    # Perform the actual edits
+    add_edit_server_form = AddEditServerForm(self.driver)
+    add_edit_server_form.editTestServer(details_modal, ip, name, private_key,
```
blueandgold: Just a couple of naming nits:

There is a couple of `editTestServer()` in different places, which would cause someone (me) to look at things closer to figure out how they are different.  I would recommend the one in the server form to just be called "editServer()".

Less important, do you think that we should just call `add_edit_server_form` to just be a more general `server_form`?
```python
from server_form import ServerForm
server_form = ServerForm(self.driver)
server_form.editTestServer(details_modal, ip, name, private_key, public_key)
```

tests/ui/add_edit_server_form.py
```diff
@@ -0,0 +1,117 @@
...
+from selenium.webdriver.support import expected_conditions as EC
+from selenium.webdriver.support.ui import WebDriverWait
+
+from layout import UfOPageLayout
+
+class AddEditServerForm(UfOPageLayout):
+
+  """Add and edit server form methods."""
+
+  def addTestServer(self, containing_element, ip, name, private_key,
```
blueandgold: nit: just `addServer()` here?

blueandgold:
Always great to see tests!  :+1:

PR #139, opened 2016-05-09
Swap out Username Field for Email for Admins

xwsxethan:
This would normally have been a DB migration, but it won't since the way the field is used currently isn't valid for an email address. For all of our standing servers, the existing username field is used as just username. No actual email addresses are present. I could append the domain to make them into email fields, but since I will actually be using the email field soon with changes for recovery, it seems more appropriate to force these (very few) accounts to just be recreated instead. Another alternative would be to keep the username and add an email address, but that seems unnecessary in this case.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/139)


xwsxethan:
Just to give some more explanation, this is the first step towards implementing a password recovery mechanism with emailing out a new password. This change was so big though, I just pulled it out to do it first. It should only encompass changing the username field to email pretty much everywhere. See above for why I'm changing the DB schema without a migration (read: it's hosed and will require resetting anyway).

blueandgold:
:+1: okay, I guess?  I can't be sure it would be more advantageous to have maintain both username and email, but if you have both, it seems like we can trade some extra space now, to save some parsing down the road.  (i.e. extract username from email)

Out of curiosity, I look at the user account in Google Directory does, and there's both username & primaryEmail fields.
https://developers.google.com/admin-sdk/directory/v1/guides/manage-users#get_all_users

xwsxethan:
Yeah, I considered keeping both. The problem is, we then have to provide a way to add an email address to the admin account. That's not too hard, but just extra work that we wouldn't get any benefit from down the road since after this initial change, we wouldn't allow an admin account to ever be setup without an email address. So to that end, if I have to modify the underlying DB manually, then I might as well just reset it to a clean state so I don't accidentally leave one user unmodified or in a weird state. I'll just recreate accounts for anyone who needs them, but using the email field this time. Luckily this is prior to release so we aren't breaking anyone.

PR #143, opened 2016-05-16
Fix queued job not running.

blueandgold:
My heroku dev instance somehow got hosed where no enqueued job would run.  Doing a flushall on the redis db helped to expose this error:
`No handlers could be found for logger “apscheduler.executors.default”`
which the commented SO thread ended up solving the problem.

The curious thing is that this happened only to my instance but not on nightly.  But this change should be not be harmful.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/143)

### DIFF
--- a/clock.py
+++ b/clock.py
@@ -2,6 +2,8 @@
 
 https://devcenter.heroku.com/articles/clock-processes-python
 """
+import logging
+
 # Use the default import to avoid module AttributeError (http://goo.gl/YM7kyZ)
 from apscheduler.schedulers.blocking import BlockingScheduler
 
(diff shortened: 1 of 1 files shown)


xwsxethan:
👍 once the tests pass.

blueandgold:
I think I know what happened here.

(1) redis db did get hosed, so the enqueued jobs are stuck; wiping the redis db with flushall did fix it so that future enqueued job will get processed; i documented this in the dev guide

(2) i was testing with a brand new instance without any data populated; thus the job failed; the scheduler attempts to log the failure, but it couldn't because it needs the python logging library which was not initialized here; so this PR really fixed this here

PR #144, opened 2016-05-17
JSON Security and Content Type to Block XSS

xwsxethan:
The changes here were to add the requested content type and charset to the JSON responses along with a prefix to mitigate XSS. Each change ended up being fairly small in terms of lines, but they had to be copy and pasted across several different places.

Each JSON handler needed to add the prefix and headers as necessary. The prefix then has to be stripped in the frontend when parsing. To remove it, iron-ajax elements actually have a specific property that can be set to automatically strip the prefix. Iron-form elements contain an iron-ajax element (under the form's request property), so they can be set similarly. Doing this for dynamically generated forms is a little more tricky, but I've managed to work it out.

The other changes were to update the unit tests to account for the JSON prefix when checking responses. The functional tests all passed when rerunning them without any modifications, but feel free to check on your own as necessary.

Finally, the change to the error handler was more in depth since it was using flask.jsonify instead of flask.response. Everywhere else uses flask.response, so I modified that and updated the associated frontend handlers since the format became slightly different. All of the JSON handlers should now function identically.

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/144)

### DIFF
--- a/ufo/__init__.py
+++ b/ufo/__init__.py
@@ -37,11 +37,7 @@ whooshalchemy.whoosh_index(app, models.User)
 whooshalchemy.whoosh_index(app, models.ProxyServer)
 
 JSON_HEADERS = {'Content-Type': 'application/javascript; charset=utf-8'}
-JSON_PREFIX_FOR_HTML = ")]}'"
-# JSON_PREFIX_FOR_HTML is the value that gets stripped in the client side, not
-# JSON_PREFIX since handling the extra new line character is difficult.
-# JSON_PREFIX is the value that gets prefixed server side.
-JSON_PREFIX = JSON_PREFIX_FOR_HTML + "\n"
+JSON_PREFIX = ")]}'\n"
 
 @app.after_request
 def checkCredentialChange(response):
--- a/ufo/__init__.py
+++ b/ufo/__init__.py
@@ -36,6 +36,13 @@ from ufo.database import models
 whooshalchemy.whoosh_index(app, models.User)
 whooshalchemy.whoosh_index(app, models.ProxyServer)
 
+JSON_HEADERS = {'Content-Type': 'application/javascript; charset=utf-8'}
+JSON_PREFIX_FOR_HTML = ")]}'"
+# JSON_PREFIX_FOR_HTML is the value that gets stripped in the client side, not
+# JSON_PREFIX since handling the extra new line character is difficult.
+# JSON_PREFIX is the value that gets prefixed server side.
+JSON_PREFIX = JSON_PREFIX_FOR_HTML + "\n"
+
 @app.after_request
 def checkCredentialChange(response):
   """Save credentials if changed"""
--- a/ufo/__init__.py
+++ b/ufo/__init__.py
@@ -36,8 +36,14 @@ from ufo.database import models
 whooshalchemy.whoosh_index(app, models.User)
 whooshalchemy.whoosh_index(app, models.ProxyServer)
 
+# The headers and prefix listed below are to help guard against XSSI. The
+# prefix specifically causes us to escape out of any client that attempts to
+# execute the JSON as code. We don't use any callbacks or functions in our
+# returned JSON, but the prefix would catch it by causing execution to stop if
+# so. The prefix is supplied in the resource dictionaries so that it can be
+# stripped away on the client side when making AJAX calls.
 JSON_HEADERS = {'Content-Type': 'application/javascript; charset=utf-8'}
-JSON_PREFIX = ")]}'\n"
+XSSI_PREFIX = ")]}'\n"
 
 @app.after_request
 def checkCredentialChange(response):
--- a/ufo/handlers/admin.py
+++ b/ufo/handlers/admin.py
@@ -21,7 +21,8 @@ def admin_list():
     A json object with 'items' set to the list of admins in the db.
   """
   items_dict = {'items': models.AdminUser.get_items_as_list_of_dict()}
-  return flask.Response(json.dumps((items_dict)), mimetype='application/json')
+  return flask.Response(ufo.JSON_PREFIX + json.dumps((items_dict)),
+                        headers=ufo.JSON_HEADERS)
 
 @ufo.app.route('/admin/add', methods=['POST'])
 @ufo.setup_required
--- a/ufo/handlers/admin.py
+++ b/ufo/handlers/admin.py
@@ -21,7 +21,7 @@ def admin_list():
     A json object with 'items' set to the list of admins in the db.
   """
   items_dict = {'items': models.AdminUser.get_items_as_list_of_dict()}
-  return flask.Response(ufo.JSON_PREFIX + json.dumps((items_dict)),
+  return flask.Response(ufo.XSSI_PREFIX + json.dumps((items_dict)),
                         headers=ufo.JSON_HEADERS)
 
 @ufo.app.route('/admin/add', methods=['POST'])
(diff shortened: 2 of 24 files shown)


blueandgold:
+1 Really nice work.  I had to catch up on XSSI, and I just want to point out that it's slightly different from XSS.

> #### JSON Security and Content Type to Block XSS
> 
> The changes here were to add the requested content type and charset to the JSON responses along with a prefix to mitigate XSS. Each change ended up being fairly small in terms of lines, but they had to be copy and pasted across several different places.
> 
> 
> 
> Each JSON handler needed to add the prefix and headers as necessary. The prefix then has to be stripped in the frontend when parsing. To remove it, iron-ajax elements actually have a specific property that can be set to automatically strip the prefix. Iron-form elements contain an iron-ajax element (under the form's request property), so they can be set similarly. Doing this for dynamically generated forms is a little more tricky, but I've managed to work it out.
> 
> 
> 
> The other changes were to update the unit tests to account for the JSON prefix when checking responses. The functional tests all passed when rerunning them without any modifications, but feel free to check on your own as necessary.
> 
> 
> 
> Finally, the change to the error handler was more in depth since it was using flask.jsonify instead of flask.response. Everywhere else uses flask.response, so I modified that and updated the associated frontend handlers since the format became slightly different. All of the JSON handlers should now function identically.
> 
> 
> 
> 
> 
> </details>


---

Review status: 0 of 23 files reviewed at latest revision, 3 unresolved discussions.

---

*[ufo/__init__.py, line 39 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/144#-KHzdawM6cI3A07U1A2n:-KHzdawM6cI3A07U1A2o:1796749705) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/12fcf8b5a3ef1a6c41abe0bc2adfd087f49c5378/ufo/__init__.py#L39)):*
> ```Python
> whooshalchemy.whoosh_index(app, models.ProxyServer)
> 
> JSON_HEADERS = {'Content-Type': 'application/javascript; charset=utf-8'}
> ```

nit: I think that this causes the response to be treated as JSONP?  Can we please put a comment there, and explain it's to guard against xssi?  maybe also also some benefit to call this "JSONP_HEADERS"?

---

*[ufo/__init__.py, line 40 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/144#-KHzdPHvOkTUDwL74zvH:-KHzdPHvOkTUDwL74zvI:-2062393192) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/12fcf8b5a3ef1a6c41abe0bc2adfd087f49c5378/ufo/__init__.py#L40)):*
> ```Python
> 
> JSON_HEADERS = {'Content-Type': 'application/javascript; charset=utf-8'}
> JSON_PREFIX = ")]}'\n"
> ```

nit: there might some benefit to name this as "xssi_prefix" so that we know what it's for.

---

*[ufo/static/userAddTabs.html, line 139 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/144#-KHzd-WvW11_NSiK8hOR:-KHzd-WwOr6o8MRhKnIF:-598660158) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/12fcf8b5a3ef1a6c41abe0bc2adfd087f49c5378/ufo/static/userAddTabs.html#L139)):*
> ```HTML
>         var fixedJsonText = detail.request.xhr.response;
>         var prefixIndex = fixedJsonText.indexOf(this.resources.jsonPrefix);
>         if (prefixIndex >= 0) {
> ```

nit: not that critical, but can this is in a lot of places.  Is is possible to put them somewhere common?  I can't think of any easy way to do this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/144#-:-KHzdxl35rPRjTGLNlGJ:-829216149)*

xwsxethan:
Reviewed 23 of 23 files at r1.

---

*[ufo/__init__.py, line 39 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/144#-KHzdawM6cI3A07U1A2n:-KI-HMKmlQHlsgZOQqZw:-1586300532) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/12fcf8b5a3ef1a6c41abe0bc2adfd087f49c5378/ufo/__init__.py#L39)):*
> nit: I think that this causes the response to be treated as JSONP?  Can we please put a comment there, and explain it's to guard against xssi?  maybe also also some benefit to call this "JSONP_HEADERS"?</details>
I'm not exactly sure if this qualifies as JSONP, since I don't know that it is executed. I think the padding is specifically to prevent things that attempt to execute it from actually causing any harm. If this had an actual function call or something similar, it would protect against it. Regardless, I've added a comment to help explain this further.

---

*[ufo/__init__.py, line 40 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/144#-KHzdPHvOkTUDwL74zvH:-KI-JDCvQtdcImCjGJ0l:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/12fcf8b5a3ef1a6c41abe0bc2adfd087f49c5378/ufo/__init__.py#L40)):*
> nit: there might some benefit to name this as "xssi_prefix" so that we know what it's for.</details>
Done.

---

*[ufo/static/userAddTabs.html, line 139 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/144#-KHzd-WvW11_NSiK8hOR:-KI-JEj4UssXRAu_0Dza:1857975111) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/12fcf8b5a3ef1a6c41abe0bc2adfd087f49c5378/ufo/static/userAddTabs.html#L139)):*
> nit: not that critical, but can this is in a lot of places.  Is is possible to put them somewhere common?  I can't think of any easy way to do this.</details>
Yeah this is something on my backlog to clean up. I think I may be able to put some of the common code like this into the globalHelperFunctions.js file and reference it from Polymer. It is a lot of duplication, and we have some similar ones for opening and closing different dialogs that can be refactored to something cleaner. I'm not worrying about it for now though just because of the other work we have to do if that's alright?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/144)*

blueandgold:
Review status: 8 of 23 files reviewed at latest revision, 3 unresolved discussions.

---

*[ufo/static/userAddTabs.html, line 139 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/144#-KHzd-WvW11_NSiK8hOR:-KI-KSVeDq7s7Tbo_CTP:-1387697961) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/12fcf8b5a3ef1a6c41abe0bc2adfd087f49c5378/ufo/static/userAddTabs.html#L139)):*
> Yeah this is something on my backlog to clean up. I think I may be able to put some of the common code like this into the globalHelperFunctions.js file and reference it from Polymer. It is a lot of duplication, and we have some similar ones for opening and closing different dialogs that can be refactored to something cleaner. I'm not worrying about it for now though just because of the other work we have to do if that's alright?</details>
Please submit, no blocker from me.  :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/144)*

PR #145, opened 2016-05-17
Adding a Content Security Policy to the HTML Pages

xwsxethan:
This change went a little bit down the rabbit hole, so here's the explanation for all the additional stuff beyond the actual CSP tag.

In order to limit our CSP to not allow unsafe-inline with scripts, we needed to load our JS separate from our HTML. Since our HTML gets pulled into the DOM via imports, the vulcanized JS looks like it is all inline scripts. To move these out to a file, we could simply leave them in their unvulcanized form, but that's not very clean and removes all the benefits from vulcanization. Instead, we can use a tool called crisper that allows us to have a separate vulcanized JS file in addition to our HTML. All the JS now lives in that file, and the vulcanized HTML file calls to load that script, so we don't even have to add a new import manually.

In order to actually use crisper, we need to remove any dependency on scripts being executed inline outside of Polymer. There's one place where this happens with setting a paper-dialog-scrollable element's parent dialog. This needs to happen only when the paper-dialog-scrollable is attached to the DOM after we parse out the response from a user get (when we are getting a listing of users to potentially add via checkboxes). To remove this script, I had to make a custom element that gets attached at the same time and fires its attached listener to actually set the parent dialog for the paper-dialog-scrollable. From there, I needed to add some CSS to make the dialog still display properly when rendered in this fashion.

The end result is a more secure CSP than allowing unsafe-inline and a cleaner overall setup for the scrollable dialog. Sorry for bundling this all together, but hopefully it makes sense.

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/145)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,6 @@ venv/*
 *.pyc
 ufo/static/bower_components/*
 ufo/static/vulcanized.html
+ufo/static/vulcanized.js
 node_modules/*
 chrome_driver/*
--- a/package.json
+++ b/package.json
@@ -2,7 +2,8 @@
   "name": "ufo-management-server",
   "dependencies": {
     "bower": "^1.7.1",
-    "vulcanize": "1.14.8"
+    "vulcanize": "1.14.8",
+    "crisper": "2.0.2"
   },
   "scripts": {
     "postinstall": "bower install && ./vulcanize.sh"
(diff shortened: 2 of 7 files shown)


blueandgold:
Sounds like a lot work, thanks for taking care of this!  :+1:

PR #151, opened 2016-05-24
Prevent shell injection in key distributor.

blueandgold:
Can't use shlex as that is only for python 3.  But pipes is the equivalent in python 2.7
https://docs.python.org/2/library/pipes.html

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/151)

### DIFF
--- a/ufo/services/key_distributor.py
+++ b/ufo/services/key_distributor.py
@@ -1,6 +1,7 @@
 """The module for distributing user keys to proxy servers."""
 
 from Crypto.PublicKey import RSA
+import pipes
 from rq import Queue
 import string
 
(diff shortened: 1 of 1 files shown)


xwsxethan:
:lgtm:

> #### Prevent shell injection in key distributor\.
> 
> Can't use shlex as that is only for python 3.  But pipes is the equivalent in python 2.7
> 
> https://docs.python.org/2/library/pipes.html
> 
> 
> 
</details>


---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/151#-:-KIZeYYhzUWxxDcHRIfh:1417088721)*

PR #156, opened 2016-05-26
Making OAuth Configuration Handle Required Attributes Properly

xwsxethan:
The domain and oauth code should only be required after initial setup (since those would be the only things you could change after initial setup). On initial setup, only an admin email and password is require, which disappears afterwards.

Also, moving the oauth/admin configuration up to the top of the setup page since it should happen before adding users/servers.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/156)


xwsxethan:
@blueandgold to check as well.

blueandgold:
:+1:

xwsxethan:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/156)*

PR #159, opened 2016-05-26
Add some logging to troubleshoot the blank logout dialog.

blueandgold:
After looking at the code, we are setting the resources into jinja env once before the first request.  It's possible that this request didn't succeed on the attempt, thus the globals are not set, and also not checked again on subsequent requests.  If reloading the page creates a new flask instance, then this might explain why the problem goes away as a new before_first_request() would be triggered.  It's hard to be sure as there's no logs around this.  So, let's start by doing that, and trace where this might be happening so that we can implement the right fix.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/159)


xwsxethan:
:+1:

PR #161, opened 2016-05-27
Adding Securing the Server with TLS Section to README

xwsxethan:
A corresponding section was added to the developer guide and DD.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/161)


blueandgold:
+1 QQ: Have we tested nginx ourselves to make sure our server works after installing & configuring it?

xwsxethan:
I haven't tested it myself. @jpevarnek have you tested configuring this to verify that it does in fact work?

xwsxethan:
Since we can perform nginx configuration ourselves without impacting users who want to deploy this outside of Heroku, I've taken that approach instead in the following PR:
https://github.com/uProxy/ufo-management-server-flask/pull/175

The documentation will updated there to reflect the new configuration so that this can be closed.

PR #167, opened 2016-05-27
Updating public_key Field to host_public_key Since the Backend Expects host_public_key Now

xwsxethan:
We just missed this when we changed it before, and it came up when I ran functional tests earlier. Easy fix once I figured out what was wrong.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/167)


blueandgold:
:+1:

PR #169, opened 2016-05-28

xwsxethan:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/169)


xwsxethan:
It looks like the web driver tests did not get triggered due to the bracket [ ] format around expressions. I think I need double brackets instead and possibly parenthesis. Going to reject this one.

PR #170, opened 2016-05-31
Applying Quick UI Fixes from Darrell's Review

xwsxethan:
Cleaning up some text on the add flow to make things more clear. 
- Renamed each tab in the add user section to more directly specify what to do in the given flow (as in search for users in group, search for individual users, or view all user in the domain).
- Renamed 'keys' to email address to make it more obvious what goes in a given input box.
- Added margin between search again button and results.
- Added support for no results found to show when no results were found for a group/individual user.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/170)


blueandgold:
👍

PR #171, opened 2016-05-31
Updating Bash Expression Syntax for Web Driver Triggering

xwsxethan:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/171)


blueandgold:
:lgtm: 

> #### Updating Bash Expression Syntax for Web Driver Triggering
</details>


---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[vulcanize.sh, line 39 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/171#-KJ6sTyeoNt72bYrm3y4:-KJ6sTyeoNt72bYrm3y5:-2130987035) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/14bad0c2c7eebf1ac8c21f84551f3195751dbb02/vulcanize.sh#L39)):*
> ```Shell
>   # We use set -e to make sure this will fail if the command returns an error
>   # code.
>   if [[ -d  "$STATIC_DIR" ]]; then
> ```

Not a bash expert, but what's the difference between single [ ] and double [[ ]] ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/171#-:-KJ6sQRyNCPyv4ta03R6:980077423)*

xwsxethan:
Reviewed 2 of 2 files at r1.

---

*[vulcanize.sh, line 39 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/171#-KJ6sTyeoNt72bYrm3y4:-KJ6tncA7GSR58fiFTgZ:1706045844) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/14bad0c2c7eebf1ac8c21f84551f3195751dbb02/vulcanize.sh#L39)):*
> Not a bash expert, but what's the difference between single [ ] and double [[ ]] ?
</details>
For bash, it seems that double brackets [[ ]] cause more predictable behavior by causing special characters to be parsed as bash builtins rather than terminal keywords. For example, [ a < b ] tries to redirect b to a whereas [[ a < b ]] checks if a is less than b as you would expect when writing an if statement. More generally, I was just following this example to make the if expression get parsed correctly: http://stackoverflow.com/questions/14964805/groups-of-compound-conditions-in-bash-test

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/171)*

xwsxethan:

---

*[vulcanize.sh, line 39 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/171#-KJ6sTyeoNt72bYrm3y4:-KJ6uv8TKCjE4hhA21Kf:124012379) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/14bad0c2c7eebf1ac8c21f84551f3195751dbb02/vulcanize.sh#L39)):*
> For bash, it seems that double brackets [[ ]] cause more predictable behavior by causing special characters to be parsed as bash builtins rather than terminal keywords. For example, [ a < b ] tries to redirect b to a whereas [[ a < b ]] checks if a is less than b as you would expect when writing an if statement. More generally, I was just following this example to make the if expression get parsed correctly: http://stackoverflow.com/questions/14964805/groups-of-compound-conditions-in-bash-test
</details>
In this case it doesn't really make a difference, but I was just trying to be consistent between files. Whereas in web_driver.sh, each expression in the if statement needs to be double brackets [[ ]] to be parsed correctly, or else it throws too many arguments. The addition of parenthesis there is to make it more obvious as to what is happening.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/171)*

PR #172, opened 2016-06-01

xwsxethan:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/172)


xwsxethan:
The web driver tests failed. It seems to be related to the bug we were seeing with the resource dictionaries not being loaded. Sauce labs shows the page loaded, but no text displayed in the admin dropdown, and the server section is not rendered correctly. It looks like we need to fix that before our tests will pass against master.

PR #175, opened 2016-06-01
nginx Configuration for Heroku

xwsxethan:
This adds nginx configuration during deployment on Heroku. The readme was updated specifying how this works, but basically it only gets triggered by the deployment steps on Heroku (via the Procfile and gunicorn) which means that standalone instances of the management server off of Heroku won't be impacted. I've added the 301 redirect to the HTTPS version and the HSTS header that will inform browsers if a client who has visited HTTPS gets downgraded back to HTTP somehow (man in the middle attack for instance). I forked the public nginx buildpack into uProxy so we can control exactly what goes into it and gated access similarly to this repo. You can view that repo here: https://github.com/uProxy/nginx-buildpack

This was tested on the nightly branch to ensure it works as intended on Heroku.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/175)

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,3 @@
-web: gunicorn ufo:app --log-file=-
+web: bin/start-nginx gunicorn -c gunicorn.conf ufo:app --log-file=-
 worker: python worker.py
 clock: python clock.py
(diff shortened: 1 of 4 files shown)


xwsxethan:
This worked so going to revert the other PR and update the documentation here so this can be approved in one go. Will follow up tomorrow.

blueandgold:
Super, this is fantastic! LGTM

PR #177, opened 2016-06-02
I18n Setup

xwsxethan:
This is part 1 in a series of changes that will enable us to use i18n on our front end. The overall process is as follows:
1. Flatten the tree of resource dictionaries to one level using unique keys for each value.
2. Combine the resource dictionaries into a single, large dictionary holding all resources and update the UI to utilize it alone.
3. Separate the large resource dictionary out into one dictionary just for i18n messages and one dictionary for urls, regexes, id, numbers, etc. to be used when generating the UI.
4. Refactor the i18n messages dictionary out into a separate JSON file and determine how to set it. I'm not sure yet whether we should continue using jinja environment globals for these or just make it into a separate JSON call. Making it a global reduces the amount of AJAX calls, but bloats the environment overall which may be the cause of the other problems we've been seeing with dictionaries occasionally not loading. I think AJAX offers us more flexibility here, but I'm open to suggestions.

This process will NOT handle refactoring the JS code to eliminate duplication there. Unfortunately, more duplication may be introduced as a result of making resource dictionaries flat and use unique keys (this is especially noticeable with the list-items element which has been copied into two separate elements, user-list and server-list to avoid doing if/case statements all over it).

However, I do plan to address this in the future. I've been looking at using Polymer behaviors to pull common JS code out into shared components so that all the method duplication can be reduced. See here for some [documentation](https://www.polymer-project.org/1.0/docs/devguide/behaviors) on how that could work.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/177)


ufo/services/resource_provider.py
```diff
@@ -256,63 +256,47 @@ def _get_user_resources():
...
+                                       'a valid group email address or unique '
+                                       'id.'),
+    'userAddTabId': 'userAddTab',
+    'userAddTab': 'Search for Individual Users',
+    'userAddFormId': 'userAdd',
+    'userAddSearchButton': 'Search for Individual User',
+    'userAddEmailAddressLabel': 'Email Address',
+    'userAddInputName': 'user_key',
+    'userAddEmailAddressDefinition': ('To add individual users, please provide'
+                                      ' a valid email address or unique id.'),
```
blueandgold: nit: ditto

blueandgold:
- Please make sure the webdriver tests pass.
- No preference to whether we use ajax or not to request the resources (advantage of the app being small).  Just that seems to be more work vs setting the jinja globals.

PR #179, opened 2016-06-03
Fix issues with the invite url

jpevarnek:
Should be /invite instead of /connect, username should be getter unless
server has been reconfigured, ssh key should be under key instead of
pass.

This won't work with the current version of the client yet...coming soon

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/179)


xwsxethan:
You'll need to update the unit test for that function (https://github.com/uProxy/ufo-management-server-flask/blob/master/ufo/handlers/user_test.py#L290), but otherwise looks good to me. :+1: once that passes.

blueandgold:
For extra credit, the method doc needs to be updated.  Probably easier to just remove the actual description of the invite code format, since it's clear in the code that's generating it.

> You'll need to update the unit test for that function (https://github.com/uProxy/ufo-management-server-flask/blob/master/ufo/handlers/user_test.py#L290), but otherwise looks good to me. :+1: once that passes.
</details>


---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/179#-:-KJfVrfPlx765tKjIW4g:1048355543)*

PR #180, opened 2016-06-06
Making Chrome Policy Download Use XSRF via Post Instead of XSSI Prefix and Use Constant Time String Comparison for Passwords

xwsxethan:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/180)


blueandgold:
:+1:  It's really good that you found a 3rd party lib for the constant time compare.

PR #188, opened 2016-06-07
Making Dialog Cancellations Reset Contained Forms

xwsxethan:
This makes the user and server add forms along with server edit not track state when exited. When these forms are modals, it is easy to force them to go through a single close dialog function where we can simply wipe the state. However, when allowing click outside to cancel or hitting the escape key to cancel, this fires a separate event. By setting a listener for that event and calling the appropriate close function, we achieve the same overall behavior.

I updated the close flow for the user add tabs and form elements to streamline this process. Now a close registered in any form (group add, individual add, or domain add) bubbles up to the tabs element which then resets each sub form and closes the modal. This is overall nicer than each form having a means of resetting each other form along with itself before closing.

Similarly in the add modal element, the listener for the cancel event gets delegated down to either the user add tabs to close (achieving the same flow as above) or the server add form to close.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/188)

### DIFF
--- a/ufo/static/addModal.html
+++ b/ufo/static/addModal.html
@@ -38,7 +38,6 @@
         },
       },
       listeners: {
-        'iron-form-error': 'handleFormError',
         'iron-overlay-canceled': 'handleCancelled',
       },
       ready: function() {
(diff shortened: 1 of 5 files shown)


blueandgold:
Cool, awesome to figure this out!

PR #191, opened 2016-06-08
Updating Bower JSON File to Hopefully Reduce Travis Bower Failures

xwsxethan:
We've been seeing a lot of bower failures in Travis CI recently when attempting to run tests. These used to be fairly rare and easily fixable by just telling Travis CI to re-execute the test, in which case it would succeed to install the necessary bower dependencies and proceed as expected. However, these have increased dramatically over the past week.

Since the only change that seemed to have an impact on this was updating Polymer to version 1.5, that seems to be the culprit. My guess is that other components updated minor versions which required the new Polymer. Those minor versions were acceptable to bower but the major change to Polymer 1.5 was not, causing a conflict. I can't understand why this behavior was intermittent, so perhaps this assumption is wrong. In any case updating to Polymer 1.5 has fixed the issue

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/191)


blueandgold:
:+1:  Have our webdriver tests pass with this change?

xwsxethan:
Locally yes, but the ones from Sauce Labs have been failing for a variety of reasons. They won't even get that far currently because bower fails so often. I'm hoping this will fix that so we can try to debug the Sauce Labs tests more closely.

PR #193, opened 2016-06-08
Adjust When XSRF Token is Looked Up for Rendering Improvement

xwsxethan:
This moves the xsrf token itself to the base HTML template and to the top of the body so it only needs to be specified in one place then get picked up by custom elements as they load.

We were calling a function to lookup the token as soon as an input element within any custom element was rendered, which may be before the actual token is rendered. For example, the admin dialog may get rendered before the input which will hold the XSRF token. In the admin dialog, that token is attempted to be copied to the local form before the real token is rendered, causing a problem with the entire element rendering. To alleviate this, we now lookup the real XSRF token in the ready function of each custom element, which should not be called until that element is actually ready and would need said token, long after the real token itself has been rendered.

I'm hoping this will help the issue we were seeing with elements not loading correctly when resource dictionaries fail. This may be related since in the cases where the XSRF token wasn't found, the whole element would error out and prevent further loading (thus stopping the resource dictionary potentially). I'll have to run this on nightly to verify.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/193)

### DIFF
--- a/ufo/static/chromePolicy.html
+++ b/ufo/static/chromePolicy.html
@@ -59,7 +59,10 @@
         },
       },
       ready: function() {
-        this.xsrfToken = document.getElementById('globalXsrf').value;
+        var xsrfElement = document.getElementById('globalXsrf');
+        if (xsrfElement) {
+          this.xsrfToken = .value;
+        }
       },
       enableSpinner: function() {
         this.set('loading', true);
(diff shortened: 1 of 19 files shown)


blueandgold:
👍  Good thinking, hope this does help!

PR #194, opened 2016-06-08
Fixing XSRF Element Value (Missed a Copy Paste)

xwsxethan:
I missed the actual element somehow and browser caching managed to save me locally, but the last change broke nightly. This is what it should have been.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/194)


blueandgold:
:+1:

PR #196, opened 2016-06-15
Forcing the Vulcanized JS File Not to be Cached to Prevent the UI From Becoming Broken

xwsxethan:
This makes the vulcanized JS file into a handler basically so that we can set a specific cache timeout for it. There are other ways to do this with flask, but this was the best way I could find to limit it to this single file. As best as I can tell, the problems seen on nightly and prod come from this file getting cached and executed at a different time than expected, leading things to render improperly. While the caching is not the underlying issue, this change should prevent the problem in the meantime without everyone manually disabling caching in their browser. I'm going to test on nightly after this is submitted to verify this does indeed work.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/196)

### DIFF
--- a/ufo/handlers/routes.py
+++ b/ufo/handlers/routes.py
@@ -15,6 +15,12 @@ from ufo.handlers import user
 def landing():
   return flask.render_template('landing.html')
 
+@ufo.app.route('/vulcanized')
+@ufo.setup_required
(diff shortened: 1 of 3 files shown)


blueandgold:
:+1:

PR #197, opened 2016-06-15
Allow Vulcanized JS to Get Pulled Down Without Authentication

xwsxethan:
I copied the method signature from above and forgot to remove the extra bits, causing the JS not to load on the login page.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/197)


blueandgold:
:+1:

PR #199, opened 2016-06-16
Fix Edit Server Flow and Web Driver Test

xwsxethan:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/199)

### DIFF
--- a/tests/ui/server_form.py
+++ b/tests/ui/server_form.py
@@ -7,6 +7,8 @@ from selenium.webdriver.support.ui import WebDriverWait
 
 from layout import UfOPageLayout
 
+import time
+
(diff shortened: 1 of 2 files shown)


xwsxethan:
At my check this morning for the automated tests, they were actually all passing to start with, but the edit test got hung up. It turns out that an enter/newline character in the private key edit server input caused the cancel button to be clicked, thus forgetting the state of all the edited fields and not actually editing the server. The enter key/newline character is somewhat expected here based on the format of the private key. By just stopping that enter key event in that specific input, the cancel button isn't clicked and the test will continue as normal and pass.

blueandgold:
:+1:

PR #201, opened 2016-06-16
Add Recaptcha to README and Allow Admins to Configure Their Own

xwsxethan:
I actually hadn't made this configurable before by hard coding our own site key. I moved that out to be picked up by the environment variable as well then added a section to the README to explain how to configure it.

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/201)


blueandgold:
:+1:

PR #202, opened 2016-06-17
Create messages.json File and Add Quaintous i18n Support

xwsxethan:
This PR creates a messages JSON file with all the actual text messages from the resource dictionaries (leaving out all ID's, numbers, regexes, recaptcha keys, url's, input names, and anything else that doesn't need to be i18n-ed). The resource dictionary itself is unchanged for the moment (will delete each of these from it in a followup PR).

Quaintous-i18n is a package for pulling down different translated messages JSON files and referencing the messages in that file from Polymer. I haven't changed any of the references yet, but I will in a separate PR. I broke this up to make the followup hopefully smaller as it will touch the resource dictionary and most of the Polymer components we've created.

I also added the logic on the backend to parse out what language to use based on the Accept-Language header. The parsing and mapping to an actual language prefix (such as en, es, fr, etc.) is rudimentary, but I left a TODO to update it once we know what actual languages we will use and what prefixes to support. That prefix is set into the session which the base html page then uses with quaintous-i18n in order to pull down the right JSON file.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/202)


blueandgold:
:+1:

PR #203, opened 2016-06-17
Use Quaintous-i18n for i18n Messages in Polymer Elements and Remove Messages from Resource Dictionary

xwsxethan:
A lot of Polymer elements/components are modified in this PR, but as you'll see the changes are pretty small in nature, just copied multiple times over. Each change is basically as follows:
1. Make the dom module reference a new property for the i18n message (rather than referencing resources).
2. Make the binding one way instead of two way (use square brackets [] instead of curly braces {}).
3. Set the new property inside of the attached event handler function for the dom module (have to use attached to ensure that quaintous-i18n is already loaded) using the **i18n** function from quaintous-i18n.
4. Add quaintous-i18n as a relative import. This isn't actually necessary because of how we handle vulcanization, but it is still included to show where the reference comes from.

Finally, I removed all the messages from the resource dictionary since the resource dictionary is no longer being reference for them.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/203)

### DIFF
--- a/ufo/services/resource_provider.py
+++ b/ufo/services/resource_provider.py
@@ -8,22 +8,16 @@ from ufo.services import regex
 
 
 def _get_resources():
-  """Get the resources for all UI components including messages for i18n.
+  """Get the resources for all UI components that need to be shared.
 
-    This dictionary of resources is returned as one giant blob for the time
+    This dictionary of resources is returned as one large blob for the time
     being to simplify passing all these parameters to the UI. These could be
-    separated, but the messages for i18n are easier to edit and pickup in a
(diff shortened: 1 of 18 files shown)


xwsxethan:
@blueandgold ping

blueandgold:
:+1:  Sorry this slipped under my radar.  So, thanks for the ping.

blueandgold:
Please make sure the webdriver stuff works for i18n.  :)

xwsxethan:
The web driver tests do pass. They were the first thing I check once it was all done.

PR #205, opened 2016-06-20
Hide the network jail setting until it's available.

blueandgold:
Save some head-scratching by hiding an unavailable option.  :)

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/205)

### DIFF
--- a/ufo/handlers/settings.py
+++ b/ufo/handlers/settings.py
@@ -55,11 +55,11 @@ def edit_settings():
   # Setup link. For now, I just want these settings to be edittable somewhere.
 
   config = ufo.get_user_config()
-
+  # Uncomment the network jail when it's available.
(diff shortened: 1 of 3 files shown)


blueandgold:
@xwsxethan , there's a build error on this PR.  Can you please take a look?  Thanks!

xwsxethan:
Can you double check that the functional tests pass with this removed? Specifically, just the SettingsComponentTest which I think is the only related one. If so, 👍 

As for the Travis error, if it is something in bower not pulling down a dependency, then it is a known issue. Just restart the process on the Travis page via the circle button, and it should fix itself.

blueandgold:
The webdriver test actually did find a bug in the policy update handler.  Glad that we have them!

xwsxethan:
Thanks and looks great! :+1:

PR #206, opened 2016-06-21

xwsxethan:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/206)


xwsxethan:
The test actually passed but then got confused after master got updated from a separate change. I'm just going to close this and let it run again tonight.

PR #209, opened 2016-06-22
Wrapping whoosh indexing in a try catch in case it fails for some reason.

xwsxethan:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/209)


goldspin:
+1, I like how the error message also tells the user what to try

PR #211, opened 2016-06-23
Make Vulcanized HTML not Cached to Avoid i18n Problems

xwsxethan:
There's an issue in the web driver tests with our vulcanized HTML file getting cached and executing the i18n calls early, such as before the messages.json file is loaded. Making the file not cached fixes this for now.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/211)

### DIFF
--- a/ufo/handlers/routes.py
+++ b/ufo/handlers/routes.py
@@ -15,6 +15,11 @@ from ufo.handlers import user
 def landing():
   return flask.render_template('landing.html')
 
+# There's an issue in the web driver tests with our vulcanized JS and HTML
+# files getting cached and executing the i18n calls early, such as before the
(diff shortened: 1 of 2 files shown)


ufo/handlers/routes.py
```diff
@@ -15,8 +15,12 @@
 def landing():
   return flask.render_template('landing.html')
 
-@ufo.app.route('/vulcanized')
-def vulcanized():
+@ufo.app.route('/vulcanized_html')
+def vulcanized_html():
```
blueandgold: nit: Can we add a comment here to document that the no caching is to avoid i18n problem?

blueandgold:
:+1:

PR #213, opened 2016-06-23
Align the text in the 3rd column of user/server lists with the header.

blueandgold:
Screenshot of what the fix looks like.  Please note: I tried to combine with the "third-div" css rule, but the alignment is still off.
screenshot from 2016-06-23 16 29 59

This is what it looks like before.
screenshot from 2016-06-23 16 22 53

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/213)

### DIFF
--- a/ufo/static/serverList.html
+++ b/ufo/static/serverList.html
@@ -30,8 +30,8 @@
     paper-item, paper-icon-item {
       border-top: 1px solid #e5e5e5;
     }
-    paper-button.thirdDiv {
-      /* Left align the paper-button text in the third div. */
+    /* Left align the paper-button text in the third div. */
+    paper-button.alignLeftText {
       display: inline;
     }
     #columnHeader {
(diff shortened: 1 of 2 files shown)


xwsxethan:
Since these can't both be combined with third-div, I'd suggest naming them something more clearly different, such as button-under-third-div. Also, it's odd that you couldn't use the same third-div class as the selector in the server-list element. That should work there. In the user-list element you could probably remove the new thirdDiv class and just use .third-div > paper-button (or .third-div paper-button, I can never remember if you need space or > for indirect descendants) as the selector since it is structured differently under the form. The end result looks great though, so :+1: once the selectors are cleaned up.

xwsxethan:
Sorry I realized that first comment wasn't clear. I'd suggest either renaming what you have to something more obviously different from third-div if you'd like to keep it or using the different selectors I mentioned if that makes more sense. No need to do both obviously.

blueandgold:
Done.  Basically revert the naming back to what it was before: "alignLeftText", with the comment to indicate this is for the 3rd div only.

xwsxethan:
Looks good to me 👍

PR #217, opened 2016-06-28
Adding support for versioning and showing the version in the UI.

xwsxethan:
This creates a file (version.json) that only holds the version so it can be easily queried from the UI. The base.html file was updated to point to a new element (versionDisplay) and pass in a url to query version.json from. The new element just grabs that json file and displays the value of the version field within. This allows us to simply update that one file each time we push a new version.

In the future I plan to add support for checking whether this version matches that in the public github repo and alerting admins that they may be behind master/production/yourBranchHere.

This also adds a manifest file for creating a web application. This will allow you to pin the application as an app on your phone by clicking "Add to Home Screen" when visiting it from Chrome. If this is a problem, then I can fork it off to a different PR, but it seemed somewhat similar in nature since this is all about metadata so I've left it for now.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/217)


ufo/static/manifest.json
```diff
@@ -0,0 +1,21 @@
+{
+  "background_color": "#009788",
+  "description": "A web application for controling usage of uProxy and securing access for users in your domain to distributed proxy servers.",
```
blueandgold: nit: spelling is "controlling"

blueandgold:
:+1:

PR #221, opened 2016-07-11
Updating Chrome Policy json format to match how it should be specified

xwsxethan:
Each policy must have a value attribute that gives the actual rather than just matching key to value directly.

---
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/221)

### DIFF
--- a/ufo/handlers/chrome_policy.py
+++ b/ufo/handlers/chrome_policy.py
@@ -14,6 +14,11 @@ def _make_chrome_policy_json():
 
   This policy string has the following form:
 
+  {
+    "validProxyServers": {"Value": map_of_proxy_server_ips_to_public_key},
+    "enforceProxyServerValidity": {"Value": boolean}
+  }
+
   Returns:
     A json string of current chrome policy.
   """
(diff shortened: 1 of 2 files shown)


blueandgold:
:lgtm:   Just a comment on testing this.

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[ufo/handlers/chrome_policy.py, line 18 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/221#-KMPzUp4zJL3-K8VfB2q:-KMPzUp4zJL3-K8VfB2r:-330418899) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/1f50d00bf52b0facd51383a2e938e2733a425f9b/ufo/handlers/chrome_policy.py#L18)):*
> ```Python
> 
>   {
>     "validProxyServers": {"Value": map_of_proxy_server_ips_to_public_key},
> ```

It's great to document this, but I would also feel so much better if we can have a test to verify the proper formatting is being generated/used.  It can be something as simple as looping through all the keys of the policy_dictionary, and assert that the inner key of the value is "Value".  :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/221#-:-KMQ-W5x83XkWdj3J-hi:-1788728508)*

xwsxethan:
Reviewed 1 of 1 files at r1, 1 of 1 files at r2.

---

*[ufo/handlers/chrome_policy.py, line 18 \[r1\]](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/221#-KMPzUp4zJL3-K8VfB2q:-KMQ2d6kSmviRobz6xxo:-499146009) ([raw file](https://github.com/uproxy/ufo-management-server-flask/blob/1f50d00bf52b0facd51383a2e938e2733a425f9b/ufo/handlers/chrome_policy.py#L18)):*
> It's great to document this, but I would also feel so much better if we can have a test to verify the proper formatting is being generated/used.  It can be something as simple as looping through all the keys of the policy_dictionary, and assert that the inner key of the value is "Value".  :)
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/221)*

PR #222, opened 2016-07-13
Fix i18n Reference and Attempt to Fix Functional Tests

xwsxethan:
I started this a while back to address some of the issues with i18n and timing. I believe I got the i18n part working to fix that issue, but the functional tests are having trouble with it. This addresses part of that, but the rest will come in a subsequent PR. From what I've seen, we seem to have trouble getting an element as the return value form web driver wait. By gating on the wait and then looking up the value manually, the issue seems to alleviate. There are still problems, but this gets us part of the way.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/222)


blueandgold:
Very cool.  :+1:

PR #224, opened 2016-07-21
Making the close dialog work consistently

xwsxethan:
It was failing because the reset value was coming up undefined if the user had never entered edit mode. In that case we don't need to reset anything however, so I just checked for that and returned out which allows the close process to proceed.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/224)


blueandgold:
Awesome!  :+1:

PR #225, opened 2016-07-21
Making the dropdown menu use iron pages to switch between elements.

xwsxethan:
This will remove the flickering issue that was seen when switching between flows on the dropdown. Instead of closing one paper dialog to open another, I now just have a single paper dialog and swap the content using iron pages.

I also fixed an error with the password label not showing up due to the key being mismatched. I accidentally made the l lowercase instead of uppercase, so I just went ahead and fixed it.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/225)


blueandgold:
:100:

PR #230, opened 2016-07-22
Create default value for textarea so it doesn't come up as undefined.

xwsxethan:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/230)


blueandgold:
Nice :)  :+1:

PR #231, opened 2016-07-22
Fix alignment for button text in the edit server dialog.

blueandgold:
I couldn't see why inline-block was used here.  But everything looks and works fine with this fix.

Before:
screenshot from 2016-07-22 09 10 07

After
screenshot from 2016-07-22 13 15 29

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/231)

### DIFF
--- a/ufo/static/editServerDialog.html
+++ b/ufo/static/editServerDialog.html
@@ -44,7 +44,7 @@
       display: none;
     }
     .showElement {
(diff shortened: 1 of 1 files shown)


xwsxethan:
:+1:

blueandgold:
disregard, just a noops comment

PR #236, opened 2016-07-29
Improve the ssh key instruction.

blueandgold:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/236)

### DIFF
--- a/DeploymentGuide.md
+++ b/DeploymentGuide.md
@@ -119,7 +119,7 @@ To view the logs of an instance:
 1. Know which user the ssh-client will be using to access the proxy server.  Currently, this is hard-coded to root.
 1. Either have an existing key-pair or generate it.
 1. Copy the public key to the user who the ssh-client will be accessing the proxy server.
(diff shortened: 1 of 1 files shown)


xwsxethan:
If you're going this far, would you mind also adding a link or command example for generating the ssh keys? I end up just googling this every time, and it might save others the same trouble. I'm pretty sure the command would be as follows:

`ssh-keygen -t rsa -b 4096 -C "your_email@example.com"`

In any case 👍 on the existing change.

blueandgold:
Good idea. Done.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/236#-:-KNrvGSQYCjQOnIoX_lJ:778570693)*

blueandgold:
retrigger travis

PR #242, opened 2016-08-02
Make admin password label actually show up.

xwsxethan:
I swear I fixed this before, but maybe it was a different element.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/242)


blueandgold:
:+1:

PR #247, opened 2016-08-03
Allowing recaptcha pass through with a message when it is not configured.

xwsxethan:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/247)

### DIFF
--- a/ufo/__init__.py
+++ b/ufo/__init__.py
@@ -35,7 +35,7 @@ else:
   app.logger.error('No recaptcha site or secret key found. Please configure ' +
                    'RECAPTCHA_SITE_KEY and RECAPTCHA_SECRET_KEY in the ' +
                    'environment variables.')
-  RECAPTCHA_ENABLED_FOR_APP = False
+  # RECAPTCHA_ENABLED_FOR_APP stays false
 
 # any instance-specific config the user wants to set, these override everything
 app.config.from_pyfile('application.cfg', silent=True)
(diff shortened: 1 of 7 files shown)


ufo/__init__.py
```diff
@@ -25,13 +25,17 @@
...
+MAX_FAILED_LOGINS_BEFORE_RECAPTCHA = 10
 if 'RECAPTCHA_SITE_KEY' in os.environ and 'RECAPTCHA_SECRET_KEY' in os.environ:
   app.config['RECAPTCHA_SITE_KEY'] = os.environ['RECAPTCHA_SITE_KEY']
   app.config['RECAPTCHA_SECRET_KEY'] = os.environ['RECAPTCHA_SECRET_KEY']
+  RECAPTCHA_ENABLED_FOR_APP = True
 else:
   app.logger.error('No recaptcha site or secret key found. Please configure ' +
                    'RECAPTCHA_SITE_KEY and RECAPTCHA_SECRET_KEY in the ' +
                    'environment variables.')
+  RECAPTCHA_ENABLED_FOR_APP = False
```
blueandgold: Hmmm, should this be here?  When the if block is true, RECAPTCHA_ENABLED_FOR_APP = True.  But when existing the if block, it will be false again?  Just double-checking.

blueandgold:
:+1:  Please take a look at the one comment that I made.

ufo/__init__.py
xwsxethan: No, it should be true if the variables are found and false otherwise. I set it to an initial value so that it is always defined, but I can confirm that it does still work when set to true. Plus we have tests for this.

xwsxethan: Wait, I think I see what you mean now. Since it's initially false, it doesn't need to be set to false again. Sorry, I wasn't comprehending at first. I'll remove this then.

PR #250, opened 2016-08-08
Fixing issues with the functional tests

xwsxethan:
These all passed when run locally so hoping to let the next nightly build give it a try.

The changes to the admin flow test have to do with a change we missed with how the add admin response gets shown. It is always shown, but empty when there is no response rather than not being found. The changes to the landing page test are for the tests which leave a details dialog open. Web driver seems to have trouble clicking the close button on the details dialog occasionally. By attempting that and then just reloading the page if it fails, that allows us to proceed during the tearDown phase which was when the errors actually happened.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/250)

### DIFF
--- a/tests/ui/landing_page_test.py
+++ b/tests/ui/landing_page_test.py
@@ -24,12 +24,6 @@ class LandingPageTest(BaseTest):
   def tearDown(self):
     """Teardown for test methods."""
     landing_page = LandingPage(self.driver)
-    try:
-      user_details_button = landing_page.get_element(
-          LandingPage.USER_DETAILS_BUTTON)
-      user_details_button.click()
-    except:
-      self.driver.get(self.args.server_url + flask.url_for('landing'))
     landing_page.remove_test_user(BaseTest.TEST_USER_AS_DICT['name'],
                                   self.args.server_url,
                                   should_raise_exception=False)
@@ -107,6 +101,8 @@ class LandingPageTest(BaseTest):
 
     self.assertUserCanBeDisabledAndThenEnabled(False, True)
 
+    landing_page.tryToCloseDetailsDialogAndRefreshIfFail(self.args.server_url)
+
   def testCreateNewInviteCode(self):
     """Test that creating a new invite code actually generates a new one."""
     landing_page = LandingPage(self.driver)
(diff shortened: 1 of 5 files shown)


tests/ui/landing_page_test.py
```diff
@@ -24,6 +24,12 @@ def setUp(self):
   def tearDown(self):
     """Teardown for test methods."""
     landing_page = LandingPage(self.driver)
+    try:
+      user_details_button = landing_page.get_element(
```
blueandgold: Wish I have a snapshot of this failure to see why it can not click.  Do you think we can make the get_element() more robust, such as with an expected condition?
https://github.com/uProxy/ufo-management-server-flask/blob/7bb7d38860435400f402782c7b87162d2ab950a8/tests/ui/base_driver.py#L22

tests/ui/landing_page_test.py
```diff
@@ -24,6 +24,12 @@ def setUp(self):
...
+          LandingPage.USER_DETAILS_BUTTON)
+      user_details_button.click()
+    except:
+      self.driver.get(self.args.server_url + flask.url_for('landing'))
```
blueandgold: If the landing page get reloaded here on exception, does the user_details_button still need to be clicked as above?

tests/ui/landing_page_test.py
xwsxethan: The issue was more so with the actual click portion I believe, not getting the element so I don't see the need.

tests/ui/landing_page_test.py
xwsxethan: No, if the page gets reloaded then the details dialog won't be open at all so there's no need to close it.

tests/ui/landing_page_test.py
blueandgold: If that's the case, then I would move the try statement down a line.  Also add a comment to explain sometimes click just doesn't work here.

Also, are we sure this is not a real bug, vs webdriver just not working?  i.e. do we know why the click is not working?

blueandgold:
👍

xwsxethan:
The try to click and refresh stuff was now moved over to its own function. It is only needed for a few tests on the landing page, so I just call it specifically from there instead of during every tear down, since doing so would cause a problem if it actually isn't present during another test's tear down. This should be cleaner all around.

PR #252, opened 2016-08-08
Making i18n retry after 1 second if it fails initially.

xwsxethan:
This change is really pretty straightforward in spite of its size. Basically, all the i18n text variable setting is moved out to its own function for each element. In the attached listener, that function is called, just as normal.

The trick is that after the initial call, a timeout is set for 1 second. After 1 second, the extra code will execute which just checks if the last i18n text to be set actually received an i18n-ed value (if it's value is not just its key). If it did not, then retry setting the i18n text for everything in that element. There is no subsequent retry, so if it fails twice, then it just fails until the whole page reloads (this likely means the key isn't defined or doesn't have a translation for the given language). This at least lets us attempt to recover in the event of timing issues with loading different components. This isn't ideal of course, but it's something.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/252)

### DIFF
--- a/DeveloperGuide.md
+++ b/DeveloperGuide.md
@@ -200,3 +200,15 @@ The `00 02 * * *` part is the syntax for cron which denotes that the job will ru
 This is already configured to run on a test machine. No developer should need to do this, but it is documented here for informational purposes and in case of errors for debugging.
 
 Once the pull request has been created and the tests have executed, all that remains is a final approval on the PR and the actual merge. Travis CI will maintain the results of the tests in case there is a failure, in which case a developer should reject the PR, and create a fix for whatever is broken. If all tests pass and there are no further concerns, a thumbs up can be given and the merge performed.
+
+## Assorted Notes
+
+### Retry for i18n After 1 Second
+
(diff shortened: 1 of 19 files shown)


blueandgold:
This is better than nothing.  But, what is the reason why the i18n values are not properly applied?  Just trying to understand the issue.

Also, what would happen on the first failure?  Would there be anything visible to the user?  Such as showing the i18n key, and then 1 second later it will be corrected?

xwsxethan:
It's a race condition essentially. The attached handler for each element gets fired when that element is fully rendered in the DOM. At that point, we try to call I18N to get the value for each key. If I18N doesn't have our messages yet, it just returns the same key you gave it (which leads to the key being displayed in the UI). The messages may not be there yet since they are loaded with an async call which can complete after the DOM gets rendered. This doesn't seem to happen at all locally (likely due localhost just caching them and serving them fast enough). This fallback is basically if the messages weren't there before, wait one second then try again.

blueandgold:
Okay.  It would be awesome if we can document your last comment somewhere, so that it would explain why the 1 sec retry is necessary.

blueandgold:
:+1:  Looks awesome now, especially the section on how the i18n works.

PR #257, opened 2016-08-16
Adding explicit close menu action to force it to close between adding and removing an admin.

xwsxethan:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/257)


blueandgold:
👍

PR #258, opened 2016-08-16
Make oauth configuration use AJAX to preserve errors and redirect from the client.

xwsxethan:
This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/258)


blueandgold:
👍

PR #260, opened 2016-08-17
Separating the initial admin setup step from the oauth configuration step.

xwsxethan:
This doesn't solve all our problems with the initial setup flow, but it's a first step in that direction. This just makes a separate first admin setup element and makes it the primary entry point. The next step will be to move around the oauth flow and clean up some other issues I've been seeing.

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/260)


blueandgold:
Cool!  :+1:

PR #265, opened 2016-08-18
Swap between add elements based on how many are present

xwsxethan:
This change creates an intermediate element (addListFlipper) which is given the last ajax result of users/servers. Based on the amount of users/servers, it will set an iron page to show either the initial add user/server flow (copied from the setup page) if there are currently 0 users/servers and the landing page's listing if there are more than 0 users/servers. This enables the flow we would like for onboarding so that the initial admin is prompted to add users/servers as expected from the landing page.

The setup page will now only show the initial admin configuration with everything else hidden until it is configured.

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/265)

### DIFF
--- a/ufo/services/resource_provider.py
+++ b/ufo/services/resource_provider.py
@@ -89,6 +89,10 @@ def _get_resources():
     'regexes': regex.REGEXES_AND_ERRORS_DICTIONARY,
     'jsonPrefix': ufo.XSSI_PREFIX,
     'maxFailedLoginsBeforeRecaptcha': ufo.MAX_FAILED_LOGINS_BEFORE_RECAPTCHA,
+    'userAddListFlipperId': 'userAddListFlipper',
+    'proxyServerAddListFlipperId': 'proxyServerAddListFlipper',
+    'userAddTabsId': 'userAddTabs',
+    'proxyServerAddFormId': 'serverAddFormHolder',
   }
 
 def set_jinja_globals():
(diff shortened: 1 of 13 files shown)


blueandgold:
Looks good overall.  I think that this is a pretty significant new behavior that it would be awesome to have a webdriver test for it.

---

Review status: 0 of 13 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/265#-:-KPTO-pgLp2X0lEmqWad:b-92z2mc)*

xwsxethan:
Yeah probably so, but that will be a significant undertaking itself. The way the web driver tests are setup, they are meant to not really alter what you have configured. For instance, it always adds new users then tries to remove those rather than removing users you actually have. Similarly for admins and proxy servers. Triggering this would involve removing all the users/proxy servers you have. If we're okay with the tests actually having the ability to do that, then I can work on adding them in the future. Otherwise, this might have to be left to be verified manually.

---

Reviewed 13 of 13 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/265#-:-KPTPRU-T0GU0saHIwmw:b-rab4rs)*

blueandgold:
:lgtm:  Please feel free to merge.  Just please either add a TODO or a bug to just track this.  Thanks.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/uproxy/ufo-management-server-flask/265#-:-KPTRwWh8_Aw2yhGJHff:b-b80c5a)*

PR #274, opened 2016-09-12
Update Deployment Guide to Create Backup Deploy Button

xwsxethan:
The deploy to heroku button isn't working due to github's content source policy. I'm adding a backup so we have another link that works even if the button doesn't render.

---

This change is [Reviewable](https://reviewable.io/reviews/uproxy/ufo-management-server-flask/274)


blueandgold:
👍

